PR #12, opened 2017-06-21
Features/modal

pciavald:
- linter not passing because of properties typed as `any`, I tried to change to the real data type but then the variable doesn't accept all function calls
- e2e testing not passing because it complains that `FirstTimeModalComponent` is not registered in any ngModule, while it is registered in the `ModalModule` which is loaded by `AppModule`


rynomster:
for the e2e testing, you need to import the module into the test that is failing. will have a look at the tests / linter

PR #13, opened 2017-06-22
RPC: transactions

kewde:
Added new RPC data that includes blind tx, which also helped identify an issue.
Added support for stealth addresses
Fixed staking reward

Prepared getParameters() for usage with RPC

### DIFF
--- a/src/app/overview/overview.component.html
+++ b/src/app/overview/overview.component.html
@@ -27,6 +27,7 @@
         </ul>
       </app-header>
       <transaction-table
+        amountOfTxsToDisplay=5
         [displayHeader]="false"
         [displayInternalHeader]="true"
         [displayPagination]="false"
--- a/src/app/shared/shared.module.ts
+++ b/src/app/shared/shared.module.ts
@@ -14,7 +14,8 @@ import { TableComponent } from './table/table.component';
 @NgModule({
   imports: [
     CommonModule,
-    AccordionModule
+    AccordionModule,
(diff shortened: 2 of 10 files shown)


rynomster:
There's an issue with the pagination.. When you switch to page 2, you can't go back to page 1.
I think we also need to add a limit for the overview page, which seems to be showing all the transactions from page 1.
Have a look at the slice pipe.. Also, we won't have all the transactions in memory, we will only have 1 page of transactions in memory. We need to make sure that we have a transaction count. The transaction count on getwalletinfo doesn't seem to count the outputs created in the genesis block. We should probably include this in the filtertransactions API

kewde:
@rynomster I wasn't aware of that.

The getParameters() function doesn't return the right values just yet. I have found this out by doing the addressbook service.

I will be moving the MAX_TXS_PER_PAGE to something that can be defined when creating the component, so we can have multiple values for overview etc..

kewde:
There's an issue with using `listtransactions` combined with `txcount` (returned by `getwalletinfo`). 

For example, getwalletinfo returns a txcount = 1.
However, that one transaction has 2 outputs and 1 input belonging to yourself (balance transfer or just a tx to yourself), then listtransactions will have 2 entries..

Then there is also a difference between just normal txs and balance transfers,
a normal tx will have two entries in listtransactions (one send and one receive).

A balance transfer from normal -> blind will have two entries (two receive entries with a 'fromself' tag).

PR #15, opened 2017-06-30
WIP: addresstable

kewde:


kewde:
filteraddresses -1 returns the address count, and I can get them by owned/sendto which means this service can do both the receive tab and the address book. 

We'll need a RPC command to add, edit and delete labels and addresses.

PR #16, opened 2017-07-01
RPC: receive tab connection

kewde:


kewde:
It seems that I can't push to this repo anymore for some odd reason.
Do not forget: push fix for travis

kewde:
Ready for review and testing and possibly merging.

TODO:
display balance by address - needs rpc change or hackish fix.

rynomster:
I am merging this PR for now, but when there are no addresses, it shows "Empty label" and "Empty address" on the receive table. I think it shouldn't show anything in that case.

kewde:
@rynomster that is what I initially tried, but the code would spit out a lot of errors, it couldn't be an undefined value. I might be able to leave it empty ( = '' ).

PR #21, opened 2017-07-07
HTML/CSS: add address to addressbook (PAR-121)

pciavald:
issue PAR-121 : Address book - Add address


kewde:
Moving this to branch rpc/addAddress and integrating RPC.

kewde:
@pciavald the 'add address modal' isn't used for receiving addresses, it's for the address book but nonetheless, this will work, thanks.

PR #24, opened 2017-07-08
RPC: addAddress

kewde:


rynomster:
"(do not merge, WIP)"

src/app/wallet/address-book/address-book.component.ts
```diff
@@ -7,9 +8,76 @@ import { Component, OnInit } from '@angular/core';
...
+
+    if (this.address.length === 102 && this.address.indexOf('Tet') === 0) {
+      this.appService.rpc.call(this, 'validateaddress', [this.address], this.rpc_callbackVerifyAddress);
+    }
+
+    this.validAddress = undefined;
+  }
+
+  rpc_callbackVerifyAddress(JSON: Object) {
+    this.validAddress = JSON['isvalid'];
```
rynomster: Please don't use JSON, it is a global object.

src/app/app.component.ts
```diff
@@ -100,5 +102,9 @@ export class AppComponent implements OnInit {
     this._modalsService.updateProgress(99);
   }
 
+  checkIfSyncing() {
```
rynomster: This function doesn't seem to be used?

src/app/core/rpc/peer.service.ts
```diff
@@ -39,10 +39,11 @@ export class PeerService {
   }
 
   setPeerList(JSON: Object) {
```
rynomster: Please don't make use of name `JSON`, it is a global object.

src/app/wallet/address-book/address-book.component.ts
```diff
@@ -1,15 +1,97 @@
...
+  }
+
+  verifyAddress() {
+    if (this.address === undefined || this.address === '') {
+      this.validAddress = undefined;
+      return;
+    }
+
+    const ret = false;
+    if ((this.address.indexOf('p') === 0) === false) { // does not start with p
```
rynomster: This validation won't work for mainnet.

we should do a base58 check oninput of the address as well:
```

var base58 = {
    base58Chars :"123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz",

    check: function(field)
    {
        var el = $(field);
        var value = el.val();

        for (var i = 0, len = value.length; i < len; ++i)
            if (base58.base58Chars.indexOf(value[i]) == -1) {
                el.css("background", "#E51C39").css("color", "white");
                return false;
            }

        el.css("background", "").css("color", "");
        return true;
    }
}
```

That's what happend in shadow on input.
We could also do a regular expression to validate the address with 'pP' and 'PT'

src/app/wallet/address-book/address-book.component.ts
```diff
@@ -1,15 +1,97 @@
...
+    }
+
+    this.validAddress = undefined;
+  }
+
+  rpc_callbackVerifyAddress(JSON: Object) {
+    this.validAddress = JSON['isvalid'];
+   }
+
+  checkAddress(): boolean {
```
rynomster: Why are we returning an existing variable instead of using it directly?
Maybe this shouldn't be state driven, but more functional.

src/app/wallet/address-book/address-book.component.ts
kewde: What do the mainnet addresses start with?

This base-58 validation just checks the characters are allowed, I could add it...

src/app/wallet/address-book/address-book.component.ts
kewde: So we can make it a private variable (currently public), this way only the JS can modify it..

kewde:
Do not merge yet, still needs address validation fix from #40 .

kewde:
Mergable now

pciavald:
I tried to add an address (my own) and got this error in the log, nothing in the UI : 
screenshot_20170727_203816

pciavald:
trying again with unlocked wallet : 
screenshot_20170727_210723
But the address is not shown in the UI and that error also shows nothing in the UI

kewde:
@pciavald 

The first error is a known issue for multiple components - will fix after #53.  

The second error is odd, when you restart the GUI - does the address table display the address?

PR #27, opened 2017-07-11
RPC: status (network & lock)

kewde:
Potential TODO: 

- [ ] Add onHover tooltip to display exact peer count?
- [ ] On click of lock -> pop up modal that unlocks wallet?


src/app/core/status/status.component.ts
```diff
@@ -7,9 +11,56 @@ import { Component, OnInit } from '@angular/core';
...
+        peerList => {
+          this.peerListCount = peerList.length;
+        },
+        error => console.log('StatusComponent subscription error:' + error));
+
+    this._rpc.register(this, 'getwalletinfo', null, this.rpc_walletEncryptionStatus, 'both'); ;
+  }
+
+  getIconNumber(): string {
+    if (this.peerListCount <= 0) {
```
rynomster: consider using a switch statement:

```
switch(true) {
  case this.peerListCount <=0:
    return 0;
  case this.peerListCount <=3:
    return 2;
  case this.peerListCount <=6:
    return 3;
  case this.peerListCount <=9:
    return 4;
  case this.peerListCount <=12:
    return 5;
  default:
    return 6;
}
```

src/app/core/status/status.component.ts
```diff
@@ -7,9 +11,56 @@ import { Component, OnInit } from '@angular/core';
...
+    } else if (this.peerListCount <= 6) {
+      return '4';
+    } else if (this.peerListCount === 7) {
+      return '5';
+    } else if (this.peerListCount >= 8) {
+      return '6';
+    }
+  }
+
+  getIconLock(): string {
```
rynomster: why are we using a get function to retrieve the variable? just use the variable.

src/app/core/status/status.component.ts
```diff
@@ -7,9 +11,56 @@ import { Component, OnInit } from '@angular/core';
...
+    return this.encryptionStatus;
+  }
+
+  rpc_walletEncryptionStatus(json: Object) {
```
rynomster: switch statement

src/app/core/status/status.component.ts
```diff
@@ -7,9 +11,51 @@ import { Component, OnInit } from '@angular/core';
...
   }
 
+  getIconNumber(): number {
+    switch (true) {
+      case this.peerListCount <= 0: return 0;
+      case this.peerListCount < 4: return 2;
+      case this.peerListCount < 8: return 3;
+      case this.peerListCount < 12: return 4;
+      case this.peerListCount < 16: return 5;
+      case this.peerListCount >= 16: return 5;
```
kewde: should return 6?
Maybe have default = 0 (disconnected)?

PR #31, opened 2017-07-12
PAR-118

pciavald:
- [x] amount of blocks left sometimes goes to -1 or -2 : replaced with `waiting for peers...`
- [x] estimated time left is sometimes empty : fixed division by 0 resulting in `NaN`
- [ ] last block time is NOT the last block's date but the last RPC update date
  - related : `timeDiff` is sometimes very short due to RPC calls being not regular, and because it is based on last RPC call and not actual block time, remaining time goes up and down like a yoyo
- [ ] internal block height is sometimes superior to network block height

I moved the logic from modals/syncing.component to core/status.service, this service emits data about status everytime an RPC call is retrieved from the RPC service. ModalsComponent subscribes it for the progressbar and SyncingComponent subscribes it for the data in the view. In appComponent I removed the SyncingModal logic and integrated it in the StatusService. On the sidebar, ManuallyOpened is set before opening the Syncing modal, in order to avoiding it poping when it was closed.


pciavald:
I think the travis build failed itself, is there a way to relaunch it ?

kewde:
There are not always peers, so the amount of block left can't always be calculated.. It's set to -1 if that the case.

Correct, integrating the last block time is easy tho. I will change that..

estimated time left can't be calculated at all times..

pciavald:
ETA is still behaving strange... often it is calculated and shown for 0.2sec before being unset

src/app/modals/firsttime/confirmpassphrase/confirmpassphrase.component.spec.ts
```diff
@@ -1,19 +1,37 @@
 import { async, ComponentFixture, TestBed } from '@angular/core/testing';
-
 import { FormsModule } from '@angular/forms';
+import { HttpModule } from '@angular/http';
+
+import { ElectronService } from 'ngx-electron';
+import { PeerService } from '../../../core/rpc/peer.service';
```
kewde: please remove all these imports and all the ones you added in the other component.spec.ts files

src/app/core/status/status.service.ts
```diff
@@ -0,0 +1,144 @@
...
+          this.highestBlockHeightNetwork = height;
+          if (this.totalRemainder === -1 && this.startingBlockCount !== -1) {
+            this.totalRemainder = height - this.startingBlockCount;
+          }
+        },
+        error => console.log('SyncingComponent subscription error:' + error));
+    this._rpcService.poll();
+  }
+
+  setManuallyOpened() {
```
kewde: I've ran into the same issue with the modal service, I can't open a modal and pass extra arguments to it..
Storing the state of the syncing modal in the status service doesn't feel like a really good solution to me :P

src/app/modals/firsttime/confirmpassphrase/confirmpassphrase.component.spec.ts
pciavald: Why ? I did it because unit tests were failing without

src/app/core/status/status.service.ts
pciavald: that's right, but I'm waiting for you to PR modalsService and store it in the data

src/app/modals/firsttime/confirmpassphrase/confirmpassphrase.component.spec.ts
kewde: The rpc.module encapsulates all these imports and rpc/recovery has those already covered

pciavald: OK, I'm on it.

pciavald:
So I've removed dependencies in the .spec files. I don't understand how the `master` branch can host some code that isn't compliant with the rules of this repository... I take time to make my code compliant with the unit tests, so if there's errors on the `master` branch other devs should be aware of it to avoid useless work for everyone.

pciavald:
I have updated the first comment of this PR to reflect the state of this issue. I think it's all for me, @rynomster tell me if it suits you. @kewde I let you fix the blocktime as you proposed, also it seems strange that sometimes `internalBlockHeight > networkBlockHeight` by 1.

pciavald:
spec files fixed

kewde:
> internal block height is sometimes superior to network block height

This is true, this most likely happens when we're fully synced and just staked a block..
The network block height is calculated by going through the peer list and picking the highest number available..

The getRemainder() doesn't return a negative number, this.totalRemainder however can be negative at times, so I added a check to make sure it's a positive number when doing estimations.

kewde:
>last block time is NOT the last block's date but the last RPC update date 

True, fixing this now.

PR #46, opened 2017-07-18
RPC: select address modal

kewde:


pciavald:
ng check gives the following error : 
screenshot_20170727_164310

PR #48, opened 2017-07-19
Feature/datatable

xludx:
Implementing AddressBook, refactored services, separation of rpc-service, service and component layers.


src/app/core/rpc/address-rpc.service.ts
```diff
@@ -0,0 +1,76 @@
...
+      .post(this.API_URL, postData, options )
+      .map(response => {
+        const addressCount = response.json().result;
+        this.log.d('getAddressCount, addressCount:', addressCount);
+        return new AddressCount(addressCount);
+      })
+      .catch(this.handleError);
+  }
+
+  public findAddresses(params: Array<any>): Observable<Address[]> {
```
kewde: I'm not exactly sure if this would work as intended

```
ngOnInit() {
  // get initial addresses using default settings
    this.addressesSubscription = this.addressService
      .findAddresses(this.getPaginationParams(this.addressType, this.currentPage))
      .subscribe(
        (addresses) => {
          this.addresses = addresses;
        }
      );
```

The function returns a _new_ observable each time findAddresses is called, to which you subscribe. When an update occurs to the addresses, it would have to re-call findAddresses (currently only in ngOnInit and onPageChanged) -> address table only updates when you change the page?

I think having _one_ public observables and updating that one would avoid this situation and potentially avoids having to re-subscribe onPageChanged?

There is currently no polling for addresses so the address table doesn't update, I did add it into #24, see specialPoll()

src/app/core/rpc/address-rpc.service.ts
```diff
@@ -0,0 +1,76 @@
...
+    const postData = this.rpc.getPostData('filteraddresses', params);
+    const options = this.rpc.getRequestOptions();
+    return this.http
+      .post(this.API_URL, postData, options )
```
kewde: I'm not completely fond of having the http post code duplicates in all of the ..-rpc.service.ts, but it's small and self-explanatory so it wouldn't be too bad. 
I prefer a single rpc service with callbacks, so we can have one service to deal with generic RPC errors and ZMQ triggers, but this structure is equally good.

xludx: My initial idea was to have a generic BaseRPCService which the AddressService would extend, and which would contain all the generic and common functionalities, but I couldn't figure out yet how to implement that. Also, in that case I should have splitted the AddressRpcService to AddressRpcService and AddressCountRpcService, which I didn't want to do yet.

kewde: I like the idea of extending upon a BaseRPCService :+1:

src/app/core/rpc/address-rpc.service.ts
```diff
@@ -70,7 +70,7 @@ export class AddressRpcService {
 
   private handleError (error: Response | any) {
     const errorMsg = (error.message) ? error.message : error.status ? error.status + ' - ' + error.statusText : 'Server error';
-    this.log.error('handleError: ', errorMsg);
+    this.log.er('handleError: ', errorMsg);
```
rynomster: needs 2 rrs ? :p

src/app/wallet/address-book/address-book.component.ts
```diff
@@ -1,15 +1,159 @@
...
+  }
 
   ngOnInit() {
+
+    // update the addressCount
+    this.addressCountSubscription = this.updateAddressCount();
+
+    // get initial addresses using default settings
+    this.addressesSubscription = this.addressService
+      .findAddresses(this.getPaginationParams(this.addressType, this.currentPage))
```
kewde: If we want to update the address table, we must re-call this function..
Do we inject a service in this component to tell us when we must update or do we move this higher up?

rynomster:
# BaseRPCService
- Poll on Timeout
 - Polls at interval
- Poll on Block
 - Polls based on Latest Block Subscription?
- Poll on Transaction
 - Polls based on Latest Transaction Subscription?

xludx:
I think this is old pull request propably not needed anymore

PR #51, opened 2017-07-23
sidebar transition

pciavald:
- [x] transition open / close
- [ ] initial transition happens on load / reload and should be avoided


rynomster:
I don't know how we can avoid the initial transition.. the rest is really nice. Could we add transitions on the accordion items next?

PR #60, opened 2017-08-04
Request unlock

kewde:


src/app/core/rpc/encryptionstatus.service.ts
```diff
@@ -0,0 +1,35 @@
...
+  constructor(private _rpc: RPCService) {
+
+    // register updates to encryptionStatus
+    this._rpc.register(this, 'getwalletinfo', null, this.setEncryptionStatus, 'time');
+
+    this._encryptionStatus = Observable.create(observer => this._observerEncryptionStatus = observer).publishReplay(1).refCount();
+    this._encryptionStatus.subscribe().unsubscribe();
+  }
+
+  private setEncryptionStatus(json: Object): void {
```
rynomster: Lets call it response, json is a bit vague..

src/app/core/rpc/rpc.service.ts
```diff
@@ -169,10 +177,17 @@ export class RPCService {
 
     this._callOnBlock.forEach(_call);
     this._callOnTransaction.forEach(_call);
+    this._callOnTime.forEach(_call);
 
     this._pollTimout = setTimeout(this.poll.bind(this), 3000);
   }
 
+/**
```
rynomster: comments should start after 2 line
```
  /**
    * multiline comment
    * @param <param> <why does this param exist
    * We don't care about return type in the comment
    */

File: src/app/core/status/status.component.ts [hunk 3]
```diff
@@ -21,18 +20,32 @@ export class StatusComponent implements OnInit {
...
         peerList => {
           this.peerListCount = peerList.length;
         },
-        error => console.log('StatusComponent subscription error:' + error));
+        error => this.log.er(`peerListCount, subscription error: ${error}`));
 
-    this._rpc.register(this, 'getwalletinfo', null, this.rpc_walletEncryptionStatus, 'both');
+    this._subEncryptionStatus = this._encryptionStatusService.getEncryptionStatus()
+      .subscribe(
+        encryptionStatus => {
```
rynomster: no need for block { }

src/app/modals/shared/password/password.component.ts
```diff
@@ -89,17 +90,63 @@ export class PasswordComponent {
...
+    if (this._encryptionStatusService.getEncryptionStatusState() === 'Unlocked, staking only') {
+      const password = this.password;
+      const timeout = this.unlockTimeout;
+      const that = this;
+
+      /*
+        After unlockTimeout, unlock wallet for staking again.
+      */
+      setTimeout(
+        function() {
```
rynomster: why create function() instead of () => {} notation? Lets keep to the standard... also, you can bind, which allows you to remove the that

src/app/core/rpc/encryptionstatus.service.ts
kewde: Isn't json more descriptive of the content inside?

src/app/core/rpc/rpc.service.ts
kewde: :+1:

src/app/modals/shared/password/password.component.ts
kewde: Sublime has a function that creates setTimeout functions, I've been using but I've seen the bind() method in the rpc service. I'll start using that instead :+1:

src/app/core/status/status.component.ts
kewde: :+1:

src/app/core/rpc/encryptionstatus.service.ts
rynomster: no, it is descriptive of the serialization, but its currently deserialized to an Object..

PR #62, opened 2017-08-04
Small refactor tx

kewde:


src/app/wallet/shared/transaction.service.ts
```diff
@@ -74,24 +74,30 @@ export class TransactionService {
...
+    this.log.d(`rpc_loadTransactions_success, real tx per page: ${json.length}`);
 
-    for (let i = 0; i < JSON.length; i++) {
-      const json: Object = JSON[i];
-      this.addTransaction(json);
+    if (json.length !== this.MAX_TXS_PER_PAGE) {
+      this.log.er(`rpc_loadTransactions_success, TRANSACTION COUNTS DO NOT MATCH (maybe last page?)`);
+    }
+
+    for (let i = 0; i < json.length; i++) {
```
rynomster: Use array.forEach <-- Much faster

PR #67, opened 2017-08-08
Fix for status icon tooltip flickering

demyanrogozhin:
This is fix for visual glitch when user moves pointer over status icon tooltip get shown and than start flickering and reposition itself.
Fixed by attaching tooltip to documents body instead of parent container.


rynomster:
Thanks for the commit :+1: Are you active in slack or riot?

PR #74, opened 2017-08-16
Remove modals dependency from createwallet

kewde:


pciavald:
tests are not passing, i tried a few ways to fix it but the error message is rather unclear

kewde:
Declaration was missing, added it again

pciavald:
tests still not passing @kewde

kewde:
There will be more failed tests - I re-added the password component. I think there are more components I will have to add manually.

rynomster:
modalService is now used in createWallet to set enableClose.. Once a user has gone past step one they can't close the modal until the complete or go back.

PR #77, opened 2017-08-21
Addressbook Table: change from table to flexbox grid

demyanrogozhin:
* Address-table component converted from table to flexbox layout
* Change grid component style to support vertical align of items (used in address book)


rynomster:
Thank you once again @demyanrogozhin, this is really looking better :) was horrible without the responsiveness.

PR #80, opened 2017-08-22
Feature/daemon manager

pciavald:
- only working for linux 64bit
  - other architectures (`aarch64`, `arm`, `i686`) need to be detected with node's [process.arch](http://node.readthedocs.io/en/stable/api/process/#processarch) but `process.arch` seems to be only able to recognize `arm`, `ia32` and `x64`
    - I need to know which string gets out of each architecture's `process.arch` call
  - other OSes (`windows`, `mac`) are provided as installers, so the script cannot auto-install / auto-check version
- app relaunch not working
  - giving the argument `true` to `daemonManager.init` in `main.js` does restart the app if a download happens, but then the app seems to be trapped in a restart loop and does it every 10sec. I couldn't find why.
- RPC cookie not yet implemented (forked from wrong branch, my bad)


pciavald:
Travis seems to have a problem
screenshot_20170822_072351

PR #90, opened 2017-08-30
Migration to BTC 0.14 + GUI - GUI INTEGRATION (RPC)

vikas-cis:
Migration to BTC 0.14 + GUI - GUI INTEGRATION (RPC)


rynomster:
This doesn't work.. It never returns data to the callbacks

PR #95, opened 2017-09-06
RPC cookie

pciavald:
### This pull request implements checking if a daemon is already running on the machine before launching the embedded daemon.
- created a new version of `modules/rpc/rpc.js` as `modules/rpc/rpcCookie.js` that does not subscribe to any IPC stuff, just returns cookie credentials.
   - [ ] Both files should be merged and reorganized as one module which exports 
        - a function that just returns the RPC credentials
        - function that connects IPC
- rewrote `startDaemon` function in `modules/clientBinaries/clientBinaries.js`

Still a work in progress as I've spotted a bug but didn't understand it yet. To reproduce :
```
npm run start:electron
ps aux | grep particld
```
Partgui launches but **RPC connection doesn't seem to work**.
Quit Partgui, launch `particld` the exact same way it was launched before (copy paste the output of `ps`), and run `npm run start:electron` again.
Now the embedded daemon is not being launched, because another daemon was detected by `particl-cli`'s `getinfo` command.
However, RPC connection seems to work this time, but **sync doesn't happen**.

### To sum up there are 2 remaining problems : 
- [ ] self-launching the daemon : RPC doesn't connect
- [ ] manually launched daemon : RPC connect but daemon doesn't sync

### DIFF
--- a/main.js
+++ b/main.js
@@ -15,8 +15,6 @@ const log = require('electron-log');
 log.transports.file.appName = '.particl';
 log.transports.file.file = log.transports.file.findLogPath(log.transports.file.appName).replace('log.log', 'partgui.log');
 
-require('./modules/rpc/rpc');
-
 const daemonManager = require('./modules/clientBinaries/clientBinaries');
 
 sock.connect('tcp://127.0.0.1:30000');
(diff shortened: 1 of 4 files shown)


rynomster:
This still has allot of work left.. Rather do a PR for the whole RPC task with arguments, etc

PR #98, opened 2017-09-09
Electron Daemon Integration

pciavald:
### PAR-142
- added support for options
  - for instance -testnet --dev -rpcbind=192.168.0.28
  - defaults are in main.js, for instance the testnet and mainnet ports
  - cleaned previous --dev hacky method
- daemon manager
  - fetches latest build
  - checks hashes
  - launches daemon if not already running
  - kills daemon if it was launched in main on application exit
- RPC
  - removed bitcoin-rpc dependency

#### how to pass arguments : 
- `npm run start:electron -- -testnet`
- `./partgui -testnet`


pciavald:
- [ ] allow using another binary
- [ ] use a plain rpc call to check if daemon is running instead of particl-cli

modules/rpc/rpc.js
```diff
@@ -87,26 +97,112 @@ function mkDir(dirPath, root) {
...
+** prepares `backend-rpccall` to receive RPC calls from the renderer
+*/
+function init(options) {
+
+  HOSTNAME = options.rpcbind;
+  PORT = options.port;
+
+  // This is a factory function that returns an Observable
+  function createObservable(event, method, params) {
+    let auth = getCookie(options.testnet);
```
rynomster: Auth could potentially use -rpcuser, -rpcpassword or -rpcauth.
Just need to bare that in mind..

modules/rpc/rpc.js
```diff
@@ -87,26 +97,112 @@ function mkDir(dirPath, root) {
...
+** execute RPC call
+*/
+function rpcCall (method, params, auth, cb) {
+  var postData = JSON.stringify({
+    method: method,
+    params: params,
+    id: '1'
+  });
+
+  var options = {
```
rynomster: Some of these options could be initialized upfront, for efficiency,
actually all of them except for content-length

PR #103, opened 2017-09-20
TestNet on Electron Fix

vikas-cis:
Latest Commits broke TestNet on Electron

### DIFF
--- a/main.js
+++ b/main.js
@@ -74,10 +74,10 @@ function initMainWindow(trayImage) {
   })
 
   // handle external URIs
-  mainWindow.webContents.on('new-window', (event, url) => {
-    event.preventDefault();
-    electron.shell.openExternal(url);
-  });
+  // mainWindow.webContents.on('new-window', (event, url) => {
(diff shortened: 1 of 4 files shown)


main.js
```diff
@@ -96,10 +96,10 @@ function initMainWindow(trayImage) {
   }
 
   // handle external URIs
-  mainWindow.webContents.on('new-window', (event, url) => {
-    event.preventDefault();
-    electron.shell.openExternal(url);
-  });
+  // mainWindow.webContents.on('new-window', (event, url) => {
```
rynomster: This is here to open external links.. Needs to be here

main.js
```diff
@@ -120,7 +120,8 @@ function makeTray() {
 
   // Determine appropriate icon for platform
   if (platform === 'darwin') {
-    trayImage = path.join(__dirname, 'src/assets/icons/logo.icns');
```
rynomster: We can get rid of these conditions... always use png

PR #116, opened 2017-09-28
Account creation on start

kewde:
Opens the createwallet modal box when it detects that no wallet exists.


rynomster:
please see #147, i think this should resolve the wallet encryption on create. We just need to test and confirm whether or not its what we want

PR #128, opened 2017-09-29
Disable input box when word is empty

anandsinghparihar:


anandsinghparihar:
Please review

rynomster:
Doesn't pass linter, will test it regardless

PR #147, opened 2017-10-02
 Add encrypt button on create wallet

rynomster:
Additionally warn the users that encryption is important before creating or restoring a wallet


rynomster:
@kewde Please let me know if you agree with this.

rynomster:
Added a method to stop state calls, but that's all it currently does. I think we should merge this to master, so that Kirti can do the final merge into the material branch and we can get the development branch started tomorrow. I know it isn't ideal, but the message does say you need to restart the client. Maybe we should just add a confirm dialogue, and close the wallet so that the user can re-open, same as the qt flow, then we can add the ability to restart the daemon in the development branch.

rynomster:
I'v'e now added a call to allow us to restart the daemon.
It's still quite rough, and I think it needs review.
I've merged it into the develop branch so that Kirti can merge material (There are 100s of conflicts), so we have to freeze the code base. This is the last task we can put into master.

modules/rpc/rpc.js
```diff
@@ -207,15 +97,23 @@ function init(options) {
...
+    let auth = cookie.getAuth(options);
     return Observable.create(observer => {
-      rpcCall(method, params, auth, (error, response) => {
-        if (error) {
-          observer.error(error);
-          return;
-        }
-        observer.next(response);
-      });
+      if (['restart-daemon'].includes(method)) {
```
kewde: Sounds like a bad idea to have it like rpc.call('restart-daemon')
If the code is ran in a browser then it won't pass through this code and it will return an error.

I think you're better off adding another channel to the IPC whitelist specifically for this type commands.

src/app/modals/encryptwallet/encryptwallet.component.ts
```diff
@@ -36,29 +40,35 @@ export class EncryptwalletComponent implements OnInit {
...
-            (response: any) => {
+          .subscribe(
+            response => {
+              this._rpc.toggleState(false);
               this.alertBox.open(response, 'Wallet');
-            } ,
+              setTimeout(() => {
+                this._rpc.call('restart-daemon')
+                  .subscribe(() => {
+                    this._modalsService.open('createWallet', {forceOpen: true});
```
kewde: So how is the initial workflow?
We force pop up on the encrypt modal, and when it's encrypted and restarted we open the createWallet modal?

kewde: Ahh I see, this one should be conditional, and the createWallet should only open when the encrypt modal was triggered by createwallet

modules/rpc/rpc.js
rynomster: If it runs in a browser there is no way to restart the daemon. We should handle it appropriately, but the issue isn't here, it's in the encrypt modal where it should be handled.

src/app/modals/encryptwallet/encryptwallet.component.ts
vikas-cis: fix #148

PR #152, opened 2017-10-09
Develop

rynomster:


src/app/wallet/send/send.component.html
```diff
@@ -77,7 +77,7 @@
       <input class="col-xs-4" type="number"
         placeholder="0.00000000"
         [(ngModel)]="send.amount" (input)="verifyAmount()"
-        [ngClass]="{'verify-sucess': checkAmount() === true, 'verify-error': checkAmount() === false }">
+        [ngClass]="{'verify-sucess': checkAmount() === true, 'verify-error': checkAmount() === false }" oninput="validity.valid||(value='');" step="any">
```
kewde: oninput="validity.valid||(value='');" step="any"

I think this line was copy pasted and they forgot to remove this?

PR #153, opened 2017-10-10
New gui fixes

vikas-cis:


src/app/core/rpc/rpx.service.ts
```diff
@@ -55,7 +55,8 @@ export class RPXService {
               observer.next(data);
               break;
             case 'e':
-              observer.error((<any>data).error ? (<any>data).error : data);
+              observer.error((<any>data).error ? (<any>data).error.error ? (<any>data).error.error
```
rynomster: This doesn't event make sense.. Please revert

PR #159, opened 2017-10-12
amount > 0

kewde:


kewde:
fixed in develop but not in master (needs fixing before first release)

PR #161, opened 2017-10-13
Disabled left menu when no wallet, Progress bar changes

vikas-cis:


src/app/app.component.ts
```diff
@@ -57,6 +57,9 @@ export class AppComponent implements OnInit {
     this.log.w('warn!');
     this.log.i('info');
     this.log.d('debug');
+    setTimeout(() => {
+      this.isWalletInitialized = this._modalsService.initializedWallet;
```
rynomster: This fix is invalid.. you'll have to use an observable.. The best is probably to use the state service to create some sort of state around whether or not the wallet is initialized

src/app/app.component.html
```diff
@@ -22,25 +22,25 @@
 
           <md-content layout-padding class="submenuarea">
             <md-list class="sidesubmenu">
-              <a routerLinkActive="active" routerLink="/wallet/receive">
+              <a routerLinkActive="active" [routerLink]="(!isWalletInitialized) ? null : '/wallet/receive'">
```
rynomster: This is really untidy.. Rather disable the links, that way the users get a cursor informing them the link is disabled.

You can use a css class that has
```css
    pointer-events: none;
    cursor: not-allowed;
```

src/app/wallet/send/send.component.html
```diff
@@ -174,27 +174,36 @@ <h5 fxFlex="10%"> Amount </h5>
...
-      <md-list *ngIf="send.input === 'private'">
+      <md-list *ngIf="send.input === 'anon_balance'">
         <md-list-item>
           <div class="md-line" fxFlex="100%" fxLayout.lg="row" fxlayout.lt-sm="column" fxLayout.sm="column"
                fxlayout.lt-xs="column" fxLayout.xs="column">
             <h5 fxFlex="10%"> Privacy level </h5>
 
             <div fxFlex="'column'" fxLayoutWrap>
+              <div class="privacy-labels" fxFlex="100%">
+                    <span fxFlex="" fxLayoutAlign="start center" class="privacy-label cursor-pointer"
```
rynomster: Please fix this indentation

src/app/wallet/send/send.component.html
```diff
@@ -174,27 +174,36 @@ <h5 fxFlex="10%"> Amount </h5>
...
+                          (click)="progress = 10; setPrivacy(2)">Low</span>
```
rynomster: I don't think low should ever be less than 3, infact I think the transaction will be rejected if its less than 3

rynomster: you could even set low to 4

rynomster: Maybe create one function, setPrivacy, which will set progress and set the level? Instead of doing 2 operations in the html...

src/app/app.component.html
```diff
@@ -22,25 +22,25 @@
...
                 </md-list-item>
               </a>
-              <a routerLinkActive="active" routerLink="/wallet/history">
+              <a [class.active]="isActiveRoute('/wallet/history')" routerLink="[]" (click)="changeRoute('/wallet/history')">
                 <md-list-item>
                   <md-icon>history</md-icon>
                   <span>History</span>
                 </md-list-item>
               </a>
-              <a routerLinkActive="active" routerLink="/wallet/address-book">
```
rynomster: Ok this solution is even worse. Please can we revert to the previous one

src/app/modals/modals.service.ts
```diff
@@ -164,6 +164,7 @@ export class ModalsService {
           const locked = this._rpcService.state.get('locked');
           if (locked) {
             this.initializedWallet = true;
+            this._rpcService.state.set('activeWallet', true);
```
rynomster: Please move the setting of this stake to rpc-state.service

src/app/modals/modals.service.ts
```diff
@@ -173,8 +174,10 @@ export class ModalsService {
                 // check if account is active
```
rynomster: Please move this to rpc-state.service

PR #189, opened 2017-10-23
unpin and pin the menu bar with icon

vikas-cis:


src/app/app.component.ts
```diff
@@ -80,6 +82,24 @@ export class AppComponent implements OnInit {
...
+  toggle() {
+    this.sideMenu = !this.sideMenu;
+    if (this.sideMenu) {
+      this.isPinned = true;
+    } else {
+      this.isPinned = false;
+    }
+  }
+
+  unPin(value: string) {
```
rynomster: Please write this properly, a single assign statement

rynomster: this.isPinned = value === 'enter'

src/app/app.component.ts
```diff
@@ -80,6 +82,24 @@ export class AppComponent implements OnInit {
       });
   }
 
+  toggle() {
```
rynomster: You can reuse this function for unpin.. No need to do 2 functions, also make it inline assignment

src/app/app.component.html
```diff
@@ -12,47 +15,48 @@
...
-          <md-expansion-panel-header>Wallet</md-expansion-panel-header>
+          <md-expansion-panel-header *ngIf="isPinned">Wallet</md-expansion-panel-header>
 
           <md-content layout-padding class="submenuarea">
             <md-list class="sidesubmenu">
               <a routerLinkActive="active" [routerLink]="walletInitialized ? '/wallet/receive' : null">
                 <md-list-item>
                   <md-icon fontSet="ncIcon" fontIcon="nc-simple-right"></md-icon>
-                  <span class="menu-item">Receive</span>
+                  <span *ngIf="isPinned" class="menu-item">Receive</span>
```
rynomster: This should use css and not Javascript to hide the menu item text

src/app/app.component.html
```diff
@@ -12,47 +15,48 @@
...
       </md-list>
     </md-content>
 
     <md-content layout-padding>
-      <md-accordion displayMode="flat" [multi]="'multi'" class="md-expansion-menu">
+      <md-accordion displayMode="flat" [multi]="'multi'"
+                    class="md-expansion-menu" [isCollapsed]="false">
         <md-expansion-panel #panel1 expanded="true">
-          <md-expansion-panel-header>Wallet</md-expansion-panel-header>
+          <md-expansion-panel-header *ngIf="isPinned">Wallet</md-expansion-panel-header>
```
rynomster: Use css not Javascript to hide

src/app/app.component.ts
```diff
@@ -80,6 +82,15 @@ export class AppComponent implements OnInit {
       });
   }
 
+  unPin(value: string) {
```
rynomster: This should be called toggle

src/app/app.component.html
```diff
@@ -12,47 +15,48 @@
         <a routerLinkActive="active" routerLink="/overview">
           <md-list-item>
             <md-icon fontSet="ncIcon" fontIcon="nc-grid-48"></md-icon>
-            <span class="menu-item">Overview</span>
+            <span class="menu-item" [ngClass]="{'hide-elements': !isPinned}">Overview</span>
```
rynomster: This is still doing it based on Javascript, now it's just adding a class.. Please do this with CSS only, by setting a class higher up...

src/app/app.component.ts
vikas-cis: ok

src/app/overview/widgets/stakinginfo/stakinginfo.component.html
```diff
@@ -1,36 +1,22 @@
...
-      </tr>
-      <tr>
-        <th>Stake reward</th>
-        <td>1.<small>4949327</small>&nbsp;PART</td>
-      </tr>
-    </table>
-    <!-- .staking-overview -->
-  </md-card>
-</div>
-<!-- .staking-info -->
```
allienworks: This code ^ is AFAIK the right one. Seems like bad merge or something..

rynomster: Looks like bad merge yes

src/app/app.component.ts
```diff
@@ -80,9 +84,52 @@ export class AppComponent implements OnInit {
...
+      this.unPin = this.sideMenu ? 'Hide Menu' : 'Show Menu';
+      this.resizeMenu('toggle');
+      return;
+    }
+    if (!this.sideMenu) {
+      this.isPinned = value === 'enter';
+    }
+  }
+
+  resizeMenu(type: string) {
```
rynomster: No way to do this without hack?

vikas-cis: yeah, I have tried to do with css but it didn't worked, that the reason used kind of javascript related to all the scenario

src/app/overview/widgets/stakinginfo/stakinginfo.component.html
vikas-cis: not sure, As its saying up to date to my branch lol :D

src/app/app.component.ts
kewde: It used to work in master without all of this, did you check that out?

src/app/app.component.ts
```diff
@@ -80,9 +84,52 @@ export class AppComponent implements OnInit {
...
+    const elem = document.getElementsByClassName('mat-drawer-content') as HTMLCollectionOf<HTMLElement>;
+    let left;
+    if (type === 'toggle') {
+      left = elem[0].offsetLeft === 100 ? 250 : 100;
+    } else if (type === 'resize') {
+        if (this.window.isXS) {
+          this.showNav = false;
```
rynomster: This will always be true, and left will always be 0

vikas-cis: yeah, I think that condition is not required 👍

src/app/app.component.ts
vikas-cis: yes I did check that and found that the sidemenu used was bootstrap and now it is material  one, and I don't see the same properties available in material sidemenu. Also when I toggling the sidmenu to decrease the width the weird thing is it taking that space, so resolve those issues I used the above stuff.

src/app/app.component.html
```diff
@@ -52,33 +56,33 @@
...
                 </md-list-item>
               </a>
             </md-list>
           </md-content>
         </md-expansion-panel>
       </md-accordion>
     </md-content>
-
-    <!--md-content layout-padding>
+
```
rynomster: Random Whitespace?

src/app/app.component.html
```diff
@@ -52,33 +56,33 @@
...
       </md-accordion>
     </md-content>
-
-    <!--md-content layout-padding>
+    
+    <!-- <md-content layout-padding>
       <md-accordion displayMode="flat" [multi]="'multi'" class="md-expansion-menu">
         <md-expansion-panel #panel1 expanded="true">
-          <md-expansion-panel-header>Settings</md-expansion-panel-header>
+          <md-expansion-panel-header *ngIf="isPinned">Settings</md-expansion-panel-header>
```
rynomster: I know it's commented, but let's take this ngIf out

src/app/app.component.ts
```diff
@@ -80,6 +84,23 @@ export class AppComponent implements OnInit {
       });
   }
 
+  toggle(value: string) {
+    if (value === 'toggle' && !this.window.isXS) {
+      this.sideMenu = !this.sideMenu;
+      this.unPin = this.sideMenu ? 'Hide Menu' : 'Show Menu';
```
rynomster: Why is this called unpin? Surely we can make a dynamic tooltip and get rid of this property?

src/app/app.component.html
vikas-cis: :)

PR #190, opened 2017-10-23
Material modal refactoring

anandsinghparihar:
**Some task is still in progress like** 

-  After refactoring ModalService and ModalComponent, some two-way data binding is not working like model close button not display while we creating any wallet.
-  And another thing is ModalComponent is also displayed progress bar using ModalService since I completely removed ModalService from Modal Component so it can not show that progress bar so I plan to create another component which responsible to display a progress bar. 

@rynomster Please review and share your feedback.


anandsinghparihar:
This branch is also cover some other issue : 

- https://particl.atlassian.net/browse/PAR-205
- https://particl.atlassian.net/browse/PAR-208

src/app/modals/createwallet/createwallet.component.ts
```diff
@@ -55,6 +55,7 @@ export class CreateWalletComponent {
 
   reset() {
     this._modalsService.enableClose = true;
+    this.state.set('model:fullWidth:enableClose', true);
```
rynomster: can you call it `modal` instead of model

src/app/modals/modals.component.ts
```diff
@@ -36,50 +35,31 @@ import { MdDialogRef } from '@angular/material';
...
-          this.open(message.modal, message.data);
-        }
-      }
-    );
-
-    // update progress bar blockstatus
-    this._modalService.getProgress().subscribe(
-      progress => this.updateProgress(<number>progress)
-    );
+  constructor(private _element: ElementRef,
```
rynomster: can we follow current standard:
```javascript
  constructor(
    private _element: ElementRef,
    private _resolver: ComponentFactoryResolver,
    public _dialogRef: MdDialogRef<ModalsComponent>,
    private state: StateService) {
    }
```

src/app/modals/modals.component.ts
```diff
@@ -36,50 +35,31 @@ import { MdDialogRef } from '@angular/material';
...
+              private _resolver: ComponentFactoryResolver,
+              public _dialogRef: MdDialogRef<ModalsComponent>,
+              private state: StateService) {
   }
 
   ngOnInit() {
-    this.enableClose = this._modalService.enableClose;
+    this.state.observe('model:fullWidth:enableClose')
+      .subscribe(
```
rynomster: this should be `status => this.enableClose = status);`
we can get rid of the brackets please

src/app/modals/modals.service.ts
```diff
@@ -80,9 +81,14 @@ export class ModalsService {
         if (!this.wasAlreadyOpen(modal)) {
           this.log.d(`next modal: ${modal}`);
           this.modal = this.messages[modal];
-          this.message.next({modal: this.modal, data: data});
+          dialogRef.componentInstance.open(this.modal, {data: data});
+          // this.message.next({modal: this.modal, data: data});
           this.isOpen = true;
-          this.enableClose = true;
+          dialogRef.componentInstance.enableClose = this.enableClose = true;
+          dialogRef.disableClose = this.enableClose;
```
rynomster: does this need to observe the state as well?

src/app/modals/shared/percentage-bar/percentage-bar.component.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+    })
+    .compileComponents();
+  }));
+
+  beforeEach(() => {
+    fixture = TestBed.createComponent(PercentageBarComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
```
rynomster: can we please add tests for update Progress?

src/app/modals/createwallet/createwallet.component.ts
anandsinghparihar: :D ohh

src/app/modals/modals.service.ts
anandsinghparihar: dialogRef.componentInstance.enableClose = true. is req. to set value on first time other will be remove.

PR #199, opened 2017-10-25
chore(package): update rxjs to version 5.5.1

rynomster:


rynomster:
Breaks - need to pin 5.5.0

PR #206, opened 2017-10-25
loader added on encrypt wallet, create wallet, restore wallet and few…

vikas-cis:
… design tweaks


kewde:
The RPC state service should not be used for things like this.

rynomster:
I don't see what's wrong with using the state service for this? It is global state

kewde:
The state service flattens everything, we have no clue which variables are stored in it and under which names. I don't think it's very smart to use the same state service for our own variables, it will eventually become one big mess where nobody knows what's what.

At the very least tag the variable names with something that ensures it doesn't collide with critical variables (such as those returned by the rpc).

A better approach imo would be to get rid of the flattening and use a different state for UI/global stuff.

rynomster:
okay, @vikas-cis please prefix it with `ui:` or something obvious, so that we know its not related to application state but only to do with UI state.. At a later stage we can compartmentalise the states, but I don't see it as an immediate requirement.

vikas-cis:
Sure its done :)

PR #218, opened 2017-10-30
Widget/coldstake

kewde:


src/app/core/rpc/rpc-state.class.ts
```diff
@@ -38,20 +39,40 @@ export class RPCStateClass {
...
 
-  private initWalletState() {
+  private coldStakeLoop() {
+    if (this.rpc.state.get('locked') === false) {
+      // only available if unlocked
+      this.rpc.call('walletsettings', ['changeaddress'])
+      .subscribe(
+        response => {
+        // check if account is active
+        if (response.changeaddress === 'default') {
```
rynomster: `response => this.rpc.state.set('coldstaking', response.changeaddress === 'default' ? false : !!response.changeaddress.coldstakingaddress)`

src/app/core/rpc/rpc-state.class.ts
```diff
@@ -38,20 +39,40 @@ export class RPCStateClass {
...
   private walletLockedState() {
     this.rpc.state.observe('encryptionstatus')
-      .subscribe(status => this.rpc.state
-        .set('locked', ['Locked', 'Unlocked, staking only'].includes(status)));
+    .subscribe(status => this.rpc.state
+      .set('locked', ['Locked', 'Unlocked, staking only'].includes(status)));
   }
 
-  private initWalletState() {
+  private coldStakeLoop() {
```
rynomster: Will never loop if locked?
This should possibly be called on a unlock event? Maybe subscribe to the locked state

kewde: It will loop, it just won't do anything if it encrypted.
The rpc command walletsettings sadly requires the wallet to be unlocked..

Currently I have it in a loop, reason being if you modify it from the command line or anything then it will also display.
I could check it once and remember it, but if you have multiple UI's connected to the node then it won't work properly, seems like we could do something with ZMQ there.

src/app/core/rpc/rpc-state.class.ts
kewde: Will do

src/app/core/rpc/rpc-state.class.ts
kewde: I'll see if I can subscribe to the unlock event, that's better than looping.

PR #221, opened 2017-10-30
Add Edit on Public and Private address label and Limit min and max sc…

vikas-cis:
…ale size of electron window


main.js
```diff
@@ -44,6 +44,8 @@ function initMainWindow(trayImage) {
   mainWindow = new BrowserWindow({
     width: 1280,
     height: 720,
+    minWidth: 961,
```
rynomster: Please remove these.. Some people want full-screen on UHD.. Others want mobile type view

src/app/wallet/receive/modals/add-address-label/add-address-label.component.html
```diff
@@ -7,7 +7,7 @@ <h2 md-dialog-title>Label for new {{ type }} address</h2>
   <md-dialog-actions>
     <button md-button md-dialog-close>Cancel</button>
     <button md-button [disabled]="!addLableForm.valid" md-raised-button color="primary">
-      Add Label
```
rynomster: Make use of ngIf or something, no need for dynamic text string..

src/app/wallet/receive/receive.component.ts
```diff
@@ -353,9 +353,12 @@ export class ReceiveComponent implements OnInit {
       error => this.log.er('error'));
     }
   }
-  openNewAddress(): void {
+  openNewAddress(callType: string, text: string, add: string): void {
```
rynomster: Make these optional

main.js
vikas-cis: aah ok

PR #242, opened 2017-11-02
add a popup when user copy in address and received page and also adde…

vikas-cis:
…d paste functionality on send page


src/app/wallet/send/send.component.html
```diff
@@ -73,7 +73,7 @@ <h5 fxFlex="16.6">To Account</h5>
               <md-radio-button value="anon_balance" disabled="{{ send.input === 'anon_balance' }}" fxFlex="16.6">Private</md-radio-button>
               <div fxFlex="50">
                 <span class="balance-label">Balance: </span>
-                <span class="balance-amount enable-select">{{ getBalance(send.input) }}</span>
+                <span class="balance-amount enable-select">{{ getBalance(send.output) }}</span>
```
kewde: why change from input -> output type?

vikas-cis: the balance showing same before

kewde: The balance will indeed show the same as before, but it's non-nonsensical as the input type defines the balance. Sure in the balance transfer tab they're both the same but still, doesn't make sense.

vikas-cis: if you see the modal name are different inside two cards. That the reason we need to pass the output rather then input

kewde: This used to work, indeed it's supposed to be output.

vikas-cis: thanks

kewde:
merge conflicts but other than that the code looks good

PR #243, opened 2017-11-02
Refactor main JS architecture

pciavald:
- `npm run start:electron:fast` builds and runs without sourcemaps
- `npm run build:electron:fast` builds without sourcemaps
- GUI starts faster
- updating at runtime is disabled


src/app/core/rpc/rpx.service.ts
```diff
@@ -1,19 +1,24 @@
 import { Injectable, NgZone } from '@angular/core';
 import { Observable } from 'rxjs/Observable';
 
+// multiwallet
+// import rxIpc from 'rx-ipc-electron/lib/renderer';
```
kewde: Can't do this :P

pciavald: i'm trying everything until I find something that at least works...

main.js
```diff
@@ -25,34 +27,76 @@ let options;
...
 
-  daemon.init(_initWindow);
-  options = daemon.getOptions();
-}
+  options = _options.parse();
+  initMainWindow();
+
+  ipc.init();
+
+  daemon.check()
```
rynomster: This isn't supposed to be in main.js

main.js
```diff
@@ -25,34 +27,76 @@ let options;
...
+    .then(()            => log.info('daemon already started'))
+    .catch(()           => daemonManager.init())
+    .then(()            => multiwallet.get())
+    .then(wallets       => ipc.promptWalletChoosing(wallets, mainWindow.webContents))
+    .then(chosenWallets => daemon.start(chosenWallets, () => log.info('daemon started')))
+});
+
+app.on('quit', function (event, exitCode) {
```
rynomster: This shouldn't be in main.js, we can create a new file to initialise everything

rynomster: Initialize and quit

modules/multiwallet/multiwallet.js
```diff
@@ -0,0 +1,33 @@
...
+
+exports.get = function () {
+  return new Promise((resolve, reject) => {
+
+    spawn('ls', [ getWalletsPath() ]).then(files => {
+
+      files = files.stdout.split('\n');
+      // keep only wallet.dat and wallet_xxxx.dat files
+      files = files.filter(file => /(wallet\.dat|wallet_.+\.dat)/.test(file));
+      log.info('found wallets: ' + files);
```
rynomster: Should be log.debug?

modules/rpc/cookie.js
```diff
@@ -94,13 +94,14 @@ function mkDir(dirPath, root) {
...
 function getAuth(options) {
+
   if (options.rpcuser && options.rpcpassword) {
     return options.rpcuser + ':' + options.rpcpassword;
   }
 
-  // const COOKIE_FILE = findCookiePath() + `${options.testnet ? '/testnet' : ''}/.cookie`;
-  const COOKIE_FILE = findCookiePath() + (options.testnet ? '/testnet' : '') + '/.cookie';
   let auth;
+  let testnet = options.testnet ? '/testnet' : ''
```
rynomster: Is an additional variable necessary here?

preload.js
```diff
@@ -34,23 +35,26 @@ class SafeIpcRenderer {
...
+    this.removeListener     = protect(ipcRenderer.removeListener);
     this.removeAllListeners = protect(ipcRenderer.removeAllListeners);
-    this.send = protect(ipcRenderer.send);
-    this.sendSync = protect(ipcRenderer.sendSync);
-    this.sendToHost = protect(ipcRenderer.sendToHost);
   }
 }
 
 window.ipc = new SafeIpcRenderer([
+  'multiwallet',
```
rynomster: We should prefix this with frontend or backend so we know where the channel is

preload.js
```diff
@@ -34,23 +35,26 @@ class SafeIpcRenderer {
...
   'rpc-channel',
```
rynomster: We should prefix this with frontend or backend so we know where the channel is registered when we call it

modules/rpc/cookie.js
pciavald: just for avoiding 80+ lines, should I remove it ?

rynomster: changed the previous comment

src/app/core/rpc/rpx.service.ts
```diff
@@ -23,7 +24,36 @@ export class RPXService {
...
+    window.ipc.on('rx-ipc-check-listener', (event, channel) => {
+      const replyChannel = 'rx-ipc-check-reply:' + channel;
+      if (this.listeners[channel]) {
+        event.sender.send(replyChannel, true);
+      } else {
+        event.sender.send(replyChannel, false);
+      }
+    });
+
+    window.ipc.on('front-choosewallet', (method, wallets) => {
```
kewde: TODO: comment out or move to service.

kewde: I'm seeing lots of commenting out of multiwallet code so I presume this must be as well

rynomster:
Tests don't seem to pass

Issue #214, opened 2017-10-28
"estimated time left until synced" is not being calculated correctly


kapsr:
1. I don't think you would want to display "estimated time left until synced" if its not calculated correctly. Currently, its always off. Don't display or make correct estimate.

2. When initial screen is clicked away, there is no visible indication anymore what syncing status is. I've seen an indication in the menu in earlier wireframes, however, not implemented yet.

![estimatedtimesync](https://user-images.githubusercontent.com/32437236/32138605-d4825014-bc35-11e7-83b7-77635dafd71b.jpg)

kapsr:
3. Also: The layout of this screen is off compared to the official testnet build.

PR #244, opened 2017-11-02
Estimated time left

kewde:
Fix estimated time left
Average the number out, prevent ridiculous hopping

Someone changed some stuff without knowing what it did, it was quite hard to figure this out.


kewde:
Testing instructions:

```
user@host:~/.particl$ mv testnet old-testnet
user@host:~/.particl$ rm -rf testnet/
user@host:~/.particl$ mkdir testnet
user@host:~/.particl$ cp old-testnet/peers.dat testnet/peers.dat
user@host:~/.particl$ cp old-testnet/wallet.dat testnet/wallet.dat
```

Start Electron GUI.

rynomster:
I changed it because it was using new Date() always.. was never using mediantime, then I added lastblock so that it would constantly updated based on last block.. median time will now only update every 2 seconds, so you won't have a smooth update on the percentage bar.

kewde:
> The sync screen changes every few seconds instead of every time a block comes in

There's no ZMQ, so it's always with a polling time, I can increase the polling time.

src/app/core/rpc/rpc-state.class.ts
```diff
@@ -13,30 +13,12 @@ export class RPCStateClass {
...
     this.rpc.registerStateCall('getstakinginfo', 10000);
 
-    this.lastBlockTimeState();
     this.blockLoop();
     this.walletLockedState();
     this.coldStakeHook();
     this.initWalletState();
   }
 
-  private lastBlockTimeState() {
```
rynomster: This code calls getblockchaininfo everytime it needs to... you can put this code back and the blocks will update as required

kewde:
#214

Issue #192, opened 2017-10-24
Wrong "TX Net Amount" counting


allienworks:
![](https://i.imgur.com/lvpYz3x.png)

- Sent: 10 PART
- Fee: 0.000358 PART
- Net: 10.000358 PART

Probably some mix-up with the `+/-`

Fee shouldn't be probably displayed with `-` as it might be confusing. Everybody understands that fee is "negative" payment IMO.

**Edit:** In received payments, Net Amount behaves as expected.

PR #247, opened 2017-11-03
fix net amount calculation

kewde:


kewde:
@allienworks this should fix it. I didn't test this tho, but the problem was quite clear.

src/app/wallet/shared/transaction.model.ts
```diff
@@ -54,6 +54,22 @@ export class Transaction implements Deserializable {
     return this.txid + this.getAmount() + this.category;
   }
 
+  public getNetAmount() {
+    const amount : number = +this.getAmount();
+    if (amount < 0) { // sent
+      return amount + this.fee;
```
allienworks: Sorry, but `ng serve` complains about:
```
ERROR in /srv/http/partgui/src/app/wallet/shared/transaction.model.ts (60,28): Property 'fee' does not exist on type 'Transaction'.
ERROR in /srv/http/partgui/src/app/wallet/shared/transaction.model.ts (62,30): Property 'fee' does not exist on type 'Transaction'.
```

allienworks: When fixed, Travis will stop complaining.

kewde:
#192

PR #250, opened 2017-11-03
Amount in tx table

kewde:


kewde:
Also the width of the paste button is way too large, please fix.
Ideally it would also be on the right of the textbox.

PR #257, opened 2017-11-04
Add Sign verify message  Modal 

anandsinghparihar:


rynomster:
Please rename the pull request to something descriptive

src/app/wallet/models/sign-verify-message.ts
```diff
@@ -0,0 +1,5 @@
+export class SignVerifyMessage {
```
kewde: Can we move this model inside the directory where `signature-address-modal.component.ts` resides?
The re-usability is very low for this particular model.

kewde: same thing for the other model in there, just move it to the right directory. It's not really being shared so no point of putting them in a shared dir.

anandsinghparihar: The model belongs from a particular module, so I guess it's in a correct location? (we can also add a model inside of component but a single component may be number of a model or maybe this model will use in another place as well .. right now AddressLookUpCopy model is not only used in single component its used in another component as well.   )

kewde: That's what I'm saying, this model won't be used anywhere else so we can just add it in the directory of the component itself.

I'll look into AddressLookUpCopy

anandsinghparihar: Okay thanks Kewde !!

kewde:
@anandsinghparihar merge conflicts, rebase

PR #260, opened 2017-11-05
fix Amount class

kewde:
math.floor being crazy
fix net accounting rounding error

### DIFF
--- a/src/app/shared/util/utils.ts
+++ b/src/app/shared/util/utils.ts
@@ -16,7 +16,11 @@ export class Amount {
    * 25.9 -> '25'
    */
   public getIntegerPart(): number {
-    return Math.floor(this.amount);
+    /* Math floor is crazy.
+     * Math.floor(-0.2) -> -1.2, so fix it by adding 1 if negative &
+     * Math.floor(-9) -> -9, so only add 1 if no dot exists.
+     */
+    return Math.floor(this.amount) + Number(this.amount < 0 && this.ifDotExist());
   }
 
   /**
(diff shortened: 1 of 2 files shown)


kewde:
@allienworks

src/app/wallet/shared/transaction.model.ts
```diff
@@ -67,18 +67,20 @@ export class Transaction implements Deserializable {
...
+  public getNetAmount(): number {
     const amount: number = +this.getAmountObject().getAmount();
 
     /* If fee undefined then just return amount */
     if (this.fee === undefined) {
       return amount;
     /* sent */
     } else if (amount < 0) {
-      return amount + this.fee;
+      const net = new Amount(+amount + (+this.fee)).getAmount();
```
rynomster: Can be one return, and next return looks the same?

src/app/shared/util/utils.ts
```diff
@@ -16,7 +16,11 @@ export class Amount {
    * 25.9 -> '25'
    */
   public getIntegerPart(): number {
-    return Math.floor(this.amount);
+    /* Math floor is crazy.
+     * Math.floor(-0.2) -> -1.2, so fix it by adding 1 if negative &
+     * Math.floor(-9) -> -9, so only add 1 if no dot exists.
+     */
+    return Math.floor(this.amount) + Number(this.amount < 0 && this.ifDotExist());
```
rynomster: Why not use `Math.trunc`?

kewde: That's indeed a no-brainer, fixing.

src/app/wallet/shared/transaction.model.ts
kewde: I'll smash the two lines into one, but I prefer to keep the else if structure, I'm not very fond of inline ( ? .. : ..) stuff. Makes readability and documenting it tough :/

rynomster:
My bad, I didn't see the + and the minus.. Looked like 2 identical lines

PR #261, opened 2017-11-05
fix recovery button

kewde:


src/app/app.component.html
```diff
@@ -18,36 +18,48 @@
...
                   <span class="menu-item">Address Book</span>
                 </md-list-item>
               </a>
+            </md-list>
+          </md-content>
+        </md-expansion-panel>
+      </md-accordion>
+    </md-content>
+
+    <md-content layout-padding *ngIf="walletInitialized === false">
```
rynomster: You might want to add an additional wallet after you've added a wallet I use 2 wallets...

src/app/modals/createwallet/createwallet.component.ts
```diff
@@ -39,6 +39,7 @@ export class CreateWalletComponent {
 
   @ViewChild('passphraseComponent') passphraseComponent: ComponentRef<PassphraseComponent>;
   @ViewChild('passwordElement') passwordElement: PasswordComponent;
+ @ViewChild('passwordRestoreElement') passwordRestoreElement: PasswordComponent;
```
rynomster: Fix indentation

src/app/app.component.html
kewde: Using "add account" will completely mess up the UX because we'd have to keep track of which HD seed we're on (address generation, no separation of outputs, the list goes on). I think multiwallet support will give the users exactly what they need.

PR #267, opened 2017-11-07
Create default addresses on generating wallet

kewde:
get rid of dead code, hacks
generate first addresses by default


kewde:
Please test with clean but encrypted wallet.

PR #273, opened 2017-11-08
version 1.0

kewde:
170+ commits, seems like we should get those into master.
The GUI at this point is definitely in a better state on the develop branch.

1.0 will still require
- [x] #268 
- [x] #267 
- [x] #278 
- [x] #295 

Maybe:
- [ ] #290


main.js
```diff
@@ -23,7 +23,7 @@ let mainWindow;
 let tray;
 let options;
 
-let openDevTools = false;
+let openDevTools = true;
```
rynomster: We don't want this at all...

rynomster:
We good to go?

Issue #272, opened 2017-11-08
Receive: Empty label (+barcode) displayed before GUI retrieves addresses


kapsr:
I may be a bit neurotic when it comes to these type of things. As shown below, before my addresses are retrieved after going to the Receive page, an empty label + barcode is displayed (for like 0.2 seconds). The barcode is bigger than my actual retrieved barcode, making the transition annoying. 

![flickrreceive](https://user-images.githubusercontent.com/32437236/32528657-8f5fbeca-c434-11e7-901f-33417881769a.jpg)

![after](https://user-images.githubusercontent.com/32437236/32528659-93552fc4-c434-11e7-8d80-b97000e12264.png)

PR #274, opened 2017-11-08
fix neurotic qr

kewde:
#272 
remove useless stuff


rynomster:
```
ERROR in /home/travis/build/particl/partgui/src/app/wallet/shared/address-table/address-table.component.ts (179,69): Argument of type '1000' is not assignable to parameter of type 'string'.
ERROR in /home/travis/build/particl/partgui/src/app/wallet/receive/receive.component.ts (353,76): Argument of type '1000' is not assignable to parameter of type 'string'.
```

PR #290, opened 2017-11-09
Add progress bar to coldstake widget

kewde:


src/app/overview/widgets/coldstake/coldstake.component.ts
```diff
@@ -1,27 +1,66 @@
...
 })
 export class ColdstakeComponent implements OnInit {
 
+  /*  General   */
+  private log: any = Log.create('send.component');
   private coldStakingEnabled: boolean = undefined;
 
+  private progress: Amount = new Amount(0, 2);
+  get coldstakeProgress(): number { return this.progress.getAmount() }
+
```
rynomster: many whitespace.. needs cleanup

src/app/overview/widgets/coldstake/coldstake.component.ts
```diff
@@ -1,27 +1,66 @@
...
+    this.rpc_progressLoop();
   }
 
   ngOnInit() {
   }
 
+  /** calls listunspent, then calculate progress. */
+  private rpc_progressLoop() {
+  this._rpc.call('listunspent')
```
rynomster: fix indentation

src/app/overview/widgets/coldstake/coldstake.component.ts
```diff
@@ -1,27 +1,66 @@
...
+    .subscribe(
+      (response: Array<any>) => {
+        let activeCount = 0;
+        let totalCount = 0;
+
+        response.forEach((output) => {
+          console.log(output.amount);
```
rynomster: remove log messages

src/app/overview/widgets/coldstake/coldstake.component.ts
```diff
@@ -1,27 +1,64 @@
...
+      // TODO: Handle error appropriately
+      error => this.log.er('rpc_progressLoop: listunspent failed', error));
+
+    if (this.coldstakeProgress < 100) {
+      setTimeout(this.rpc_progressLoop.bind(this), 1000);
+    }
+  }
+
   isColdStakingEnabled() {
     return this.coldStakingEnabled;
```
rynomster: this is also unnecessary.. uses more memory for no reason, as it will copy the variable into a new memory space before returning the result.. Lets make the property public rather

kewde:
@rynomster 
https://github.com/particl/partgui/blob/develop/src/app/core/rpc/rpc-state.class.ts#L56

The ui:coldstake already gets updated every time the wallet gets unlocked, so I think it's not needed

allienworks:
The progressbar itself is OK, but we found a few issues with @dasource:
- the widget doesn't auto-update itself (e.g. from deactivated -> activated etc.)
- I added the widget variant for when cold staking reaches 100% (with "disable cold staking" button and updated text)

PR #291, opened 2017-11-09
Core update 0.15.0.3

pciavald:
## bumped version file to [0.15.0.3](https://github.com/particl/partgui/blob/master/modules/clientBinaries/clientBinaries.json#L4)

The version file [on the master branch](https://github.com/particl/partgui/blob/master/modules/clientBinaries/clientBinaries.json) of this repository controls the version that is being checked by the GUI at startup to initiate daemon upgrade. This PR is intended to be merged in the **master** branch once we're ready to trigger update to 0.15.0.3 on the GUI.

When we do so : 
- `alpha` wallets will restart during the hour, changing the state of **unlocked** and **unlocked for staking** wallets to **locked**
- `1.0` wallets will not restart, but they will update on next start

**note:** the pull requests indicates a huge amount of commits right now because it is based on `develop` and targeted to `master`. Once `develop` is merged in `master`, only the JSON file will appear.


main.js
```diff
@@ -23,7 +23,7 @@ let mainWindow;
 let tray;
 let options;
 
-let openDevTools = false;
+let openDevTools = true;
```
rynomster: We don't want this...

rynomster:
Shouldn't we do this PR into develop first?

PR #297, opened 2017-11-09
Refactor main JS architecture

pciavald:
- GUI starts faster (before daemon)
  - **may lead to error modals opening and sidebar errors triggering and not leaving even when the daemon is ready**
  - modals and side errors need to be fixed before merging to master, but we can merge to develop to allow other issues going forward #342 #343 #305 #351 #354 
- improved logging
  - we'll need #344 but I think it's alright for now
- script for bumping core version
  - `npm run generate:versions`


modules/daemon/daemon.js
```diff
@@ -0,0 +1,134 @@
...
+}
+
+exports.check = function() {
+  return new Promise((resolve, reject) => {
+
+    const _timeout = rpc.getTimeoutDelay();
+    let auth = cookie.getAuth(_options.get());
+    rpc.setTimeoutDelay(150);
+    rpc.call('getnetworkinfo', null, (error, response) => {
+      rxIpc.removeListeners();
```
kewde: Hmm interesting, checking the daemon disables the listeners?

pciavald: yes, i don't know why... comes from [here](https://github.com/particl/partgui/blob/master/modules/rpc/rpc.js#L137), @rynomster ?

modules/daemon/daemon.js
```diff
@@ -0,0 +1,135 @@
...
+      exports.check()
+      .then(() => { callback(); resolve(); })
+      .catch(() => {
+        if (exitCode === 0 && retries < maxRetries) {
+          setTimeout(daemonStartup, 1000);
+        }
+      });
+
+      retries++;
+      if (exitCode || retries >= maxRetries) {
```
kewde: `retries >= maxRetries`
This is also quite weird

modules/multiwallet.js
```diff
@@ -0,0 +1,33 @@
...
+    return path.join(app.getPath('home'), '.particl');
+  } else {
+    return app.getPath('userData');
+  }
+}
+
+exports.get = function () {
+  return new Promise((resolve, reject) => {
+
+    spawn('ls', [ getWalletsPath() ]).then(files => {
```
kewde: Does this work on windows? 'ls' vs 'dir'

pciavald:
I agree, but I think it works... I'll test it tomorrow  mardi, 14 novembre 2017, 00:32AM +01:00 de Kewde  notifications@github.com :

modules/daemon/daemon.js
pciavald: I agree, but I think it works. I'll test it and maybe see why tomorrow

modules/multiwallet.js
pciavald: good catch, thanks !

modules/daemon/daemon.js
pciavald: it actually makes sense. there's 10 retries, one every second, and when we exceed that we check for `-reindex` need, if not we exit. I think `reindex` could be done each time to boot faster, but I'm also not too reassured by the fact that there's 2 recursive calls in this function that could fire in the same call so i've protected the first one with a return.

modules/multiwallet.js
pciavald: I disabled multiwallet in this PR as it will be done in another one.

pciavald:
@rynomster can you please review this PR ? A lot of it is refactoring of your own code so there may be subtle things I missed.

modules/init.js
```diff
@@ -0,0 +1,37 @@
...
+    // .then(wallets       => ipc.promptWalletChoosing(wallets, mainWindow.webContents))
+    .then(chosenWallets => daemon.start(chosenWallets, daemonStarted));
+    // TODO: activate for daemon ready IPC message to RPCService
+    // .then(()            => ipc.daemonReady(mainWindow.webContents))
+}
+
+electron.app.on('before-quit', function beforeQuit(event) {
+  event.preventDefault();
+  log.debug('received quit signal');
+  electron.ipcMain.removeAllListeners(['rpc-channel']);
```
kewde: Hmm? no usage of rxIpc?

modules/daemon/daemon.js
kewde: I noticed that there is another place where listeners get removed in an odd way (check other comment)

PR #298, opened 2017-11-09
fast commands

pciavald:
- `npm run start:electron:fast` builds and runs without sourcemaps
- `npm run build:electron:fast` builds without sourcemaps


pciavald:
Should I also provide fast commands for `dev` launches ?

PR #307, opened 2017-11-10
(win) fix of TX table overflow

allienworks:
- fixes https://github.com/particl/partgui/issues/292


allienworks:
Wait with the merge, I'll add something else really quick..

PR #311, opened 2017-11-10
bumping GUI to 0.15.0.4

pciavald:


rynomster:
will the dmg cause issues on osx?

PR #335, opened 2017-11-11
progressbar in menu

kewde:
- progress bar in menu
- test exponential moving average for time left
- get rid of blocks state usage in blockstatus


src/app/core/rpc/peer.service.ts
```diff
@@ -51,6 +54,11 @@ export class PeerService {
       error => this.log.er(`updatePeerListLoop(): getpeerinfo error ${error}`)
     );
 
+    this._rpc.call('getblockcount').subscribe(
```
rynomster: Why would you do this?

rynomster: We already have it in state..

kewde: It caused a hanging in the last few blocks, it would keep the sync modal open longer than required

kewde:
Still needs GFX fixed.

allienworks:
I've styled the progress bar, here's a [preview for the lazy](https://imgur.com/a/443Pm) – @gerlofvanek approved? :)
It's visible while it's syncing, disappears when at 100%.

Also couldn't help not to tweak the logo & sidebar a bit.

allienworks:
@rynomster can we merge?

We need this merged so that https://github.com/particl/partgui/pull/342 can follow.

PR #337, opened 2017-11-13
Refactor tx  + filtertransactions

kewde:


pciavald:
can you add a description @kewde ? just to make sure we're not working on the same things

pciavald:
What's your status en this task @kewde ? I need the final state of Transaction to finish #377 History tabs. Or do you want to implement it ?

kewde:
@pciavald seems to have made its way into the restructure_test branch.
Will be available after the big merge to dev.

PR #339, opened 2017-11-13
Add paste button into Add new address modal

anandsinghparihar:


pciavald:
Adding a paste button works, but we should fix why macos does not paste correctly

anandsinghparihar:
@pciavald I checked in macos its working fine. I can paste using cmd+v.

Issue #334, opened 2017-11-11
No Operating system notification with gui


SativaL:
Hey since i use the gui for example after an incoming stake i dont get noticed like with qt back then (a small popup in the top right popped up and got me known), why is that ?

rynomster:
We are currently working on adding notifications

pciavald:
closing in favor of PR

PR #340, opened 2017-11-13
Transaction OS-notification implementation

vikas-cis:
#334 
Transaction Received Notification


main.js
```diff
@@ -207,8 +209,31 @@ app.on('activate', function () {
...
+    appIcon: path.join(__dirname, 'src/assets/icons/notification.png'),
+    displayTime: 6000,
+    defaultStyleText: {
+      color: '#FF0000',
+      fontWeight: 'bold'
+    },
+    maxVisibleNotifications: 1
+  });
+
+  rxIpc.registerListener('rx-ipc-message', function(title, desc, params) {
```
kewde: bad name use something like `notification`

src/app/core/rpc/rpc.service.ts
```diff
@@ -337,5 +336,12 @@ export class RPCService {
     // A poll only for address changes, triggered from the GUI!
     this._callOnAddress.forEach(this._pollCall.bind(this));
   }
+
+  /** Send Notification to the backend */
```
kewde: Not in the right place, notifications in the rpc service?

src/app/core/rpc/rpx.service.ts
```diff
@@ -65,5 +65,14 @@ export class RPXService {
     });
   }
 
+  runNotification(channel: string, ...args: any[]): Observable<any> {
```
kewde: Does this work? Variable channel names should not be allowed.
Also rpx class is not the right place to put this.

kewde:
Also will conflict with #359 & #297

main.js
vikas-cis: ok

src/app/core/rpc/rpc.service.ts
vikas-cis: Rpc using rpx where all stuffs written related to communication between frontend and backend(electron), That is the reason I just added another method which called from rpc service and emit the message to main.js

src/app/core/rpc/rpx.service.ts
vikas-cis: Variable channel names should not be allowed , yes we can change it the exact name what it is using :)

allienworks:
Can't run this, I'm getting JS error: [cannot find module "electron-notify"](https://i.imgur.com/ntusjpD.png)

Module not included in `package.json`?

allienworks:
OK, so correct me if I'm wrong, but current notifications are some "3rd party" implementation of not-OS-native notifications, right?



Electon itself offers native OS notifications:
- https://electronjs.org/docs/tutorial/notifications
- https://electronjs.org/docs/api/notification

We should implement those instead, if possible. I thought that was automatic from the beginning, so sorry that we're discussing this just now.

There would be no need to style them, but mainly the notifications would blend into user's OS nicely. Let's stick to the standards here.

vikas-cis:
yes it is a third party, I used that coz it provide more options, I think we can discuss it more with Henk and Kewde :)

vikas-cis:
@HenkSwardt and @kewde any comments related to the stuff what @allienworks suggesting ?
Thanks

pciavald:
- I also think we should use native notifications in case the third party plugin decides not to work, and it's one more dependency that is not purely needed. 
- RPC Service is also clearly not the right place to store this logic, probably a NotificationsService should be created. 

@HenkSwardt we may need to define which notifications we want to set up and their message.

allienworks:
Not working for me on Linux x64 (Gnome) :(

allienworks:
Works now, looks nice - preview of Linux/Gnome notification:

vikas-cis:
@allienworks and @anandsinghparihar  I have tested and it seems not working, Even I moved to the commit where @allienworks says its working on @anandsinghparihar system (ubuntu) and still it doesn't work. I have tested it on mac and it is working fine and I showed via screenshare with @allienworks.

vikas-cis:
@allienworks , @pciavald , @HenkSwardt ,  I have just read here 
https://github.com/hokein/electron-sample-apps/tree/master/notifications
It seems sometimes it will work on linux or sometimes not :( . We need to decide rather use the module that I have implemented before or something else we need to decide

pciavald:
It will work on all standard [freedesktop](https://www.freedesktop.org/wiki/) desktops.

pciavald:
Works nicely on arch / kde / i3

pciavald:
screenshot_20171129_145215

pciavald:
Good for me. Who's still having trouble with it ?

allienworks:
Apparently @anandsinghparihar on Ubuntu has some issues. Works on Win, Mac and (some?) Linux; so let's ship this at least to the majority of users and go from there.

allienworks:
I just received 1 stake reward and the notification poped 3 times.. any idea why?

Issue #349, opened 2017-11-14
Add "Send all" button/icon to Send page


anandsinghparihar:
Suggestion: add "Send all" button/icon to Send page.

anandsinghparihar:
@kewde , please let me know API details.

pciavald:
Logging here for convenience
```
sendtoaddress "address" amount ( "comment" "comment_to" subtractfeefromamount )

Send an amount to a given address.

Requires wallet passphrase to be set with walletpassphrase call.
Arguments:
1. "address"            (string, required) The bitcoin address to send to.
2. "amount"             (numeric or string, required) The amount in BTC to send. eg 0.1
3. "comment"            (string, optional) A comment used to store what the transaction is for. 
                             This is not part of the transaction, just kept in your wallet.
4. "comment_to"         (string, optional) A comment to store the name of the person or organization 
                             to which you're sending the transaction. This is not part of the 
                             transaction, just kept in your wallet.
5. subtractfeefromamount  (boolean, optional, default=false) The fee will be deducted from the amount being sent.
                             The recipient will receive less bitcoins than you enter in the amount field.
```

HenkSwardt:
PR Pending review #358

anandsinghparihar:
new PR : https://github.com/particl/partgui/pull/510

PR #358, opened 2017-11-14
Add "Send all" button/icon to Send page

anandsinghparihar:
Add a "Send all" button/icon to Send page.


anandsinghparihar:
@allienworks could you please check send all button design !!

allienworks:
@anandsinghparihar seems working fine, thanks!

I was trying to style the button so it fits the rest. But for some reason, the base styles of the buttons aren't included in the send component (when you look, the styles for `.mat-raised-button` are manually pasted in the styles, not inherited. Can you include the button styles from Material?

src/app/wallet/send/send.service.ts
```diff
@@ -173,7 +173,7 @@ export class SendService {
   getSendParams(
     anon: boolean, address: string, amount: number, comment: string,
     narration: string, ringsize: number, numsignatures: number) {
-    const params: Array<any> = [address, amount, '', '', false];
+    const params: Array<any> = [address, amount, '', '', true];
```
kewde: Including the fee by default should _not_ be default behaviour.

allienworks:
Styled. When @kewde's requests are done, we can merge.

src/app/wallet/send/send.service.ts
anandsinghparihar: Okay, So I guess we should use a checkbox for select & unselect `Send al`l amount ..? 

if someone select >> Send all checkbox then amount field will be disable and if use unchecked `Send All` amount field will empty.  (at the same time SendService will also update true/false)

allienworks:
@anandsinghparihar have you finished kewde's change requests?

HenkSwardt:
Hi @allienworks 

Yes all changes applied.

pciavald:
closing #349

src/app/settings/settings.component.scss
```diff
@@ -1,8 +1,13 @@
-//@import '~https://fonts.googleapis.com/icon?family=Material+Icons';
-//@import '~@angular/material/prebuilt-themes/indigo-pink.css';
+/* TODO: remove this completely */
+
+/* @import '~https://fonts.googleapis.com/icon?family=Material+Icons'; */
+/* @import '~@angular/material/prebuilt-themes/indigo-pink.css'; */
 @import "./src/assets/_config"; // import shared colors etc.
 
 /* card element */
+// test
```
pciavald: Shouldn't we clear out all unnecessary comments ?

src/app/settings/settings.component.scss
```diff
@@ -47,11 +56,11 @@ button:active {
...
 
 /* Particl inputs */
-input[type="text"] {
+/*input[type="text"] {
   background-color: #fff;
   border: 1px solid #EDEDED;
   padding: 4px;
-}
\ No newline at end of file
+}*/
```
pciavald: missing EOF

src/app/wallet/send/send.component.html
```diff
@@ -16,8 +16,10 @@
         </md-tab-group>
       </div>
       <div fxFlex="23" disableTouchRipple="true" class="bottom-line" fxLayoutAlign="left center">
-        <button md-button (click)="toggleAdvanced()" class="advanced-options"><!-- *ngIf="type === 'balanceTransfer'" -->
-          <md-icon fontSet="partIcon" fontIcon="{{ (advanced) ? 'part-circle-delete-2': 'part-circle-add-2' }}"></md-icon>
+        <button md-button (click)="toggleAdvanced()" class="advanced-options">
+          <!-- *ngIf="type === 'balanceTransfer'" -->
```
pciavald: comment

src/app/wallet/send/send.service.ts
```diff
@@ -173,7 +173,7 @@ export class SendService {
   getSendParams(
     anon: boolean, address: string, amount: number, comment: string,
     narration: string, ringsize: number, numsignatures: number) {
-    const params: Array<any> = [address, amount, '', '', false];
+    const params: Array<any> = [address, amount, '', '', this.isSubstractfeefromamount];
```
pciavald: this should be an object with key/value pairs to avoid confusion

src/app/wallet/send/send.component.ts
```diff
@@ -302,4 +304,9 @@ export class SendComponent {
     document.execCommand('Paste');
   }
 
+  sendAllBlance(): void {
```
pciavald: Balance is wrongly spelled

src/app/wallet/send/send.component.html
```diff
@@ -155,8 +161,13 @@ <h5 fxFlex="15">Amount</h5>
                      placeholder="0.00000000"
                      [(ngModel)]="send.amount" (input)="verifyAmount()"
                      [ngClass]="{'verify-sucess': checkAmount() === true, 'verify-error': checkAmount() === false }"
-                     oninput="validity.valid||(value='');" step="any">
+                     oninput="validity.valid||(value='');" step="any"
+                     [disabled]="isSendAll">
             </md-form-field>
+            <md-checkbox name="send_all" [(ngModel)]="isSendAll" (click)="sendAllBlance()" class="send-all">
```
pciavald: did you make the same mistake twice ? 😮

pciavald:
I think the design of the button should be adjusted, @allienworks @gerlofvanek ?

screenshot_20171204_181113

It does not make sense to keep the label 0.00000000 over the field. It was the behaviour before this task, can you please fix it ?

screenshot_20171204_181321

@anandsinghparihar in terms of functionnality, changing the balance origin (i.e. public / blind) instead of putting the field in red, it should update the max. It is a checkbox so it means whatever balance is selected, max will be sent. :)

anandsinghparihar:
@pciavald , I think better is to have a red field as it makes things user-friendly.

so I suppose we can stick with showing red if a user enters the value which is greater then max and we force a user to enter <=max value.

That make sense?

allienworks:
@kewde why was this closed (when unmerged)?

pciavald:
@anandsinghparihar the problem is that the field goes red without the user inputing a value. If max is checked, it should just take max. What happens now is that I select Blind, Max, then Public, max is not true anymore. Max amount should be updated when selecting another balance.

allienworks:
Can't even compile now.

As this is based on really old `develop` branch, how about doing this from scratch on current `dev` branch instead?

anandsinghparihar:
@allienworks  @gerlofvanek  please comple  design related changes !!

allienworks:
@anandsinghparihar we can't. Can not compile. Also, please switch base branch to `dev`.

anandsinghparihar:
Yes I am going to create a new branch.

allienworks:
@anandsinghparihar when you create the new branch, please close this PR.

anandsinghparihar:
New PR: new PR : https://github.com/particl/partgui/pull/510

PR #359, opened 2017-11-14
Restructure

kewde:
TODO:
- [x]  Fix tests.
- [ ]  Move sidebar out of wallet.module and into core-ui.module & redo routing.

Looking at the PR won't  reveal the big picture, I suggest looking at the new folder structure I created.
https://github.com/kewde/partgui/tree/restructure/src/app

Currently it works but there is still quite some work to do..
I've prepared the new structure for the marketplace. Segregated the wallet into it's own folder, prepared the directory for multiwallet.

The core IPC & RPC logic is not part of the wallet module anymore, it is now housed into the core folder.  This module will register all the listeners, singleton services etc. We import this on the app module (root), and we change the switch between it. This will also fix the multiwallet issue where all those error modals pop-up, as we can delay the code running the wallet module (and thus modals).

I've also created a module specific for system-wide imports like all the material, flexbox stuff.
That's all in core-ui, I'm not fond of importing all the specific material components everywhere, it really clouds the import code up.

Redoing the module structure has been a long awaited change, it will allow us in the future to take more advantage of lazy load modules, which only loads things when they're needed.

There is one ugly task left to do, which I've been dreading for a while: the current sidebar.
It's a mix of bootstrap, spread over multiple directories. It's entangled with a ton of different imports, can't really make heads or tails of it anymore.

Looking at the PR won't  reveal the big picture, I suggest looking at the new folder structure I created.
https://github.com/kewde/partgui/tree/restructure/src/app

Currently it works but there is still quite some work to do..
I've prepared the new structure for the marketplace. Segregated the wallet into it's own folder, prepared the directory for multiwallet.

The core IPC & RPC logic is not part of the wallet module anymore, it is now housed into the core folder.  This module will register all the listeners, singleton services etc. We import this on the app module (root), and we change the switch between it. This will also fix the multiwallet issue where all those error modals pop-up, as we can delay the code running the wallet module (and thus modals).

I've also created a module specific for system-wide imports like all the material, flexbox stuff.
That's all in core-ui, I'm not fond of importing all the specific material components everywhere, it really clouds the import code up.

Redoing the module structure has been a long awaited change, it will allow us in the future to take more advantage of lazy load modules, which only loads things when they're needed.

There is one ugly task left to do, which I've been dreading for a while: the current sidebar.
It's a mix of bootstrap, spread over multiple directories. It's entangled with a ton of different imports, can't really make heads or tails of it anymore.


src/app/wallet/modals/encryptwallet/encryptwallet.component.ts
```diff
@@ -20,7 +21,7 @@ export class EncryptwalletComponent {
   @ViewChild('passwordElement')
   passwordElement: PasswordComponent;
 
-  constructor(private _rpc: RPCService,
+  constructor(private _rpc: RpcService,
```
rynomster: Constructor might as well be restructured to follow standards

src/app/wallet/modals/shared/percentage-bar/percentage-bar.component.spec.ts
```diff
@@ -1,9 +1,11 @@
 import { async, ComponentFixture, TestBed } from '@angular/core/testing';
 
+import { CoreModule } from '../../../../core/core.module';
+
 import { PercentageBarComponent } from './percentage-bar.component';
 import { MdDialogModule } from '@angular/material';
```
rynomster: Should be modules, services then components

src/app/wallet/overview/widgets/stakinginfo/stakinginfo.component.spec.ts
```diff
@@ -4,7 +4,7 @@ import { FlexLayoutModule } from '@angular/flex-layout';
 import { MdCardModule } from '@angular/material';
 
 import { StakinginfoComponent } from './stakinginfo.component';
-import { StateService } from '../../../core/state/state.service';
+import { StateService } from '../../../../core/core.module';
```
rynomster: Should be modules, services then components

HenkSwardt:
Removing label - "Require test cases"

@vikas-cis and @anandsinghparihar  has implemented these and handed back to @kewde for finalization of task.

PR #377, opened 2017-11-17
History tabs

pciavald:
This pull request integrates the filtertransactions API in the History tab to allow fine-grained control on the list's contents.

### DIFF
--- a/src/app/wallet/history/history.component.html
+++ b/src/app/wallet/history/history.component.html
@@ -1,18 +1,30 @@
 <app-header>
   <div class="container-fluid">
-    <!--ul class="header-nav">
-      <li [class.active]="category == 'all'"><a href="javascript:void(0)" (click)="filterByCategory('all')">All</a></li>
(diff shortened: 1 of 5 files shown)


pciavald:
waiting for #337

allienworks:
_Suggestion:_ I was thinking about this. The obvious (UI/UX) solution would be to include tabs at the top with "Received / Sent / Staked" etc. However I think it might be better to go a different way and use checkboxes or something like that (to let the user explicitly include or exclude certain type of TXs) – e.g. I might want to see only Sent & Received TXs or Sent & Staked etc. This would give the user more flexibility and options.

pciavald:
@allienworks if you agree I will only to the #PRG side of it, then you style it how you see fit ?

allienworks:
@pciavald sure, that's why I'm here ;)

allienworks:
Outdated, closing (new PR will replace this).

PR #378, opened 2017-11-18
Number of confirmations doesnt update in Overview tab fix

### LINKED ISSUE #345: Number of confirmations doesn't update in "Overview" tab
The number of confirmations displayed in each transaction lines in "Overview" (i.e. 12+) doesn't update if staying on the same page. The number of confirmations is updated when switching sections by clicking on another tab (i.e. Send) and then clicking back on "Overview". 

Is this on purpose? It would feel better if it updated in real-time especially for monitoring staking rewards and their confirmations.
selection_004

vikas-cis:
#345


allienworks:
Tested and works. But seems like it refreshes the page in a few sec interval – not the best solution IMO. If I have expanded details of some TXs, the client automatically refreshes and every TX gets collapsed again. Could this be fixed?

Still needs code review.

vikas-cis:
@allienworks it will refresh every 10sec and it will only  refresh when you are on overview and history page.

vikas-cis:
@pciavald yes I have discussed with Ryno, also suggest the same :)

vikas-cis:
More feasible way to call the api using state as discussed with Pierre and Ryno, But still we are refreshing the whole data and at that stage if user expanded the transaction then also it get refreshed whenever new confirmations update. So for this scenario I need to discuss with @rynomster

Issue #333, opened 2017-11-11
only 24 words seeds are allowed - allow 12, 15 and 18 words too


SativaL:
a mnemonic seed with less than 24 words cant be important into the new gui. (tried with 18 words from the standalone particl generator)

The "Next" button stays grey and does nothing.

rynomster:
Thanks, this is a valid bug.. I also have some 12 and 15 word mnemonics
lying around

pciavald:
moved to PR

PR #379, opened 2017-11-18
only 24 words seeds are allowed - allow 12, 15 and 18 words too fix v…

vikas-cis:
…alidation

#333


kewde:
Have you tested this?

vikas-cis:
@kewde Yes it will not disabled 12,15,18,24 while restoring the wallet

PR #380, opened 2017-11-18
bumping core to 0.15.1.0 for develop

pciavald:
This PR will bump all clients built from **develop** to 0.15.1.0 so we can test it before release.

Wallets from the official builds will **not** bump.

I will need to do more work to automatize this behaviour before next GUI release, if i'm not done with it we can simply change it manually by pasting the commented link to master.

Please rebase `develop` if you're currently working on a feature to bump core.

### DIFF
--- a/modules/daemon/daemonManager.js
+++ b/modules/daemon/daemonManager.js
@@ -13,8 +13,13 @@ const rpc = require('../rpc/rpc');
 
 let options;
 
-// should be 'https://raw.githubusercontent.com/particl/partgui/master/modules/clientBinaries/clientBinaries.json';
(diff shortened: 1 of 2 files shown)


modules/daemon/daemonManager.js
```diff
@@ -13,8 +13,13 @@ const rpc = require('../rpc/rpc');
 
 let options;
 
-// should be 'https://raw.githubusercontent.com/particl/partgui/master/modules/clientBinaries/clientBinaries.json';
-const BINARY_URL = 'https://raw.githubusercontent.com/particl/partgui/master/modules/clientBinaries/clientBinaries.json';
+// master
+// const BINARY_URL = 'https://raw.githubusercontent.com/particl/partgui/blob/develop/modules/clientBinaries/clientBinaries.json';
```
pciavald: Use this URL when pushing to master

Issue #346, opened 2017-11-14
Add more "Items per page" in History section


cryptoguard:
Currently, it is only possible to choose between 5, 10 and 20 transactions displayed per pages. Could more options be added like 50, 100, 250, etc. 

This is easier to have many transactions on the same page when monitoring staking rewards and also to quickly search for a specific transaction. Easier to scroll than have to click the next arrow for hundreds of pages. 

Alternatively, a page menu dropdown could be added to quickly jump through a lot of pages. 
![selection_005](https://user-images.githubusercontent.com/27031337/32763502-16218e5a-c8ce-11e7-9997-afbe8df5f75d.jpg)

allienworks:
I believe this is solved already and will be included in the next release.

PR #384, opened 2017-11-20
#346 Adds more per-page options in History list

anandsinghparihar:


allienworks:
Added 50, 100 and 250 items per page options. I have only 30 TXs, but tested and it works.

Not sure if that's the default behaviour, but when you select e.g. 50 TX per page, click on some other page and go back, the settings isn't kept. Do we want it like that?

anandsinghparihar:
@allienworks Default item- per-page is 10 so if you leave current (history list ) page, then default per page value will be defined.

pciavald:
@allienworks yeh, i think that's pretty common behaviour

PR #390, opened 2017-11-21
Sync modal

anandsinghparihar:
When launched and blockchain isn't synced, the syncing modal (see #356) still fires up, even though it's empty.

We need not to launch the modal at all, syncing process is shown in the sidebar menu now.

(develop branch)
issue: https://github.com/particl/partgui/issues/381

### DIFF
--- a/src/app/modals/syncing/syncing.component.html
+++ b/src/app/modals/syncing/syncing.component.html
@@ -22,7 +22,7 @@
           <span class="value">
             {{ remainder }}
             <div fxLayoutAlign="center center" *ngIf="!syncPercentage">
-              <md-spinner class="sync-spinner"></md-spinner>
+                <md-spinner class="sync-spinner"></md-spinner>
             </div>
           </span>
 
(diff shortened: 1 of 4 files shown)


pciavald:
Can you please write a description of what I should be reviewing @anandsinghparihar ? thank you

anandsinghparihar:
description updated, please check

pciavald:
I'm getting the sync modal firing

screenshot_20171123_185650

src/app/modals/syncing/syncing.component.html
```diff
@@ -1,45 +1,51 @@
-<div *ngIf="syncPercentage">
-  <div class="container">
-    <div class="content">
+<div class="container">
+  <div class="content">
+    <div class="section wallet-syncing">
+      <div class="title">Wallet is syncing</div>
+      <p>
+        Your wallet automatically synchronizes with the Particl network after a connection is established, but this
```
pciavald: From the description I though the modals should be removed ?

allienworks: We talked about this (on meeting) and decided it would be best to leave it there. Since you can't really fully use the wallet when it's syncing, it makes sense. If you don't like it, you can still close it and you'll see progress in the sidebar. If needed, by clicking the sync progress in sidebar, you can fire the sync modal again.

pciavald: alright then, let's merge this and see later how we integrate the new filtertransactions API and Transaction model

PR #391, opened 2017-11-21
Gfx/versions

EagleEye25:
added fucntion to display daemon, client current version + added variables into html to be displayed

### DIFF
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -110,7 +110,7 @@
       <!-- Versions | TODO: move to separate component? -->
       <md-content class="client-version">
         <div class="client item">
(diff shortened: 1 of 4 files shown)


src/app/app.component.html
```diff
@@ -110,10 +110,10 @@
       <!-- Versions | TODO: move to separate component? -->
       <md-content class="client-version">
         <div class="client item">
-          Client <span class="version TO_DO">1.1.0</span>
+          Client <span class="version">{{ clientVersion }}{</span>
```
allienworks: A typo? Redundant `{` at the end..?

PR #413, opened 2017-11-26
fix: pin ngx-clipboard to 8.1.1

kewde:


rynomster:
looks like this is related to them upgrading to angular 5 https://github.com/maxisam/ngx-clipboard/issues/91

PR #418, opened 2017-11-27
Error handeling of words entered

EagleEye25:
if user enteres an invalid word it will be displayed in red, if a valid word is entered, but not correct to the generated words and next is clicked an appropriate error message will be show


rynomster:
you'll have to rebase this on the develop branch.. Right now there are too many changes coming through from master

rynomster:
lets checkout a clean createwallet, I'll help you quickly

rynomster:
We could add tests to see if random words actually add the classes required

PR #434, opened 2017-11-29
Adds close action in modal cancel button 

anandsinghparihar:
- Change cancel button action type and when a user clicks on cancel modal would close.
- Cancel button work like a submit button in receive page, add address book etc

### DIFF
--- a/src/app/wallet/wallet/shared/signature-address-modal/signature-address-modal.component.ts
+++ b/src/app/wallet/wallet/shared/signature-address-modal/signature-address-modal.component.ts
@@ -1,6 +1,6 @@
 import { Component, ElementRef, OnInit, ViewChild } from '@angular/core';
 import { FormBuilder, FormGroup, Validators } from '@angular/forms';
-import { MatDialog } from '@angular/material';
+import {MatDialog, MatDialogRef} from '@angular/material';
 import { Log } from 'ng2-logger';
 
 import { SignVerifyMessage } from './sign-verify-message.model';
@@ -39,7 +39,8 @@ export class SignatureAddressModalComponent implements OnInit {
     private _rpc: RpcService,
     private flashNotification: SnackbarService,
     private formBuilder: FormBuilder,
-    private _modals: ModalsService
+    private _modals: ModalsService,
+    private dialogRef: MatDialogRef<SignatureAddressModalComponent>
   ) {
   }
 
@@ -159,5 +160,9 @@ export class SignatureAddressModalComponent implements OnInit {
   pasteAddress(): void {
     this.addressInput.nativeElement.focus();
     document.execCommand('paste');
+  }
+
+  dialogClose(): void{
+    this.dialogRef.close();
   }
 }
--- a/src/app/wallet/wallet/shared/signature-address-modal/signature-address-modal.component.ts
+++ b/src/app/wallet/wallet/shared/signature-address-modal/signature-address-modal.component.ts
@@ -162,7 +162,7 @@ export class SignatureAddressModalComponent implements OnInit {
     document.execCommand('paste');
   }
 
-  dialogClose(): void{
+  dialogClose(): void {
     this.dialogRef.close();
   }
 }
--- a/src/app/wallet/wallet/shared/signature-address-modal/signature-address-modal.component.ts
+++ b/src/app/wallet/wallet/shared/signature-address-modal/signature-address-modal.component.ts
@@ -1,6 +1,6 @@
 import { Component, ElementRef, OnInit, ViewChild } from '@angular/core';
 import { FormBuilder, FormGroup, Validators } from '@angular/forms';
-import {MatDialog, MatDialogRef} from '@angular/material';
+import { MatDialog, MatDialogRef } from '@angular/material';
 import { Log } from 'ng2-logger';
 
 import { SignVerifyMessage } from './sign-verify-message.model';
@@ -34,14 +34,12 @@ export class SignatureAddressModalComponent implements OnInit {
 
   @ViewChild('addressInput') addressInput: ElementRef;
 
-  constructor(
-    private dialog: MatDialog,
-    private _rpc: RpcService,
-    private flashNotification: SnackbarService,
-    private formBuilder: FormBuilder,
-    private _modals: ModalsService,
-    private dialogRef: MatDialogRef<SignatureAddressModalComponent>
-  ) {
+  constructor(private dialog: MatDialog,
+              private _rpc: RpcService,
+              private flashNotification: SnackbarService,
+              private formBuilder: FormBuilder,
+              private _modals: ModalsService,
+              private dialogRef: MatDialogRef<SignatureAddressModalComponent>) {
   }
 
   ngOnInit() {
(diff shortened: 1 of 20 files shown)


anandsinghparihar:
@allienworks  I need some design tweaks to hide close button shadow.
desktop screenshot

anandsinghparihar:
@allienworks  I complete that design changes as well. @pciavald this branch is ready for merge.

pciavald:
- the top right button is misplaced and redundant with the bottom `close`. Can you delete it ?
screenshot_20171206_165130

pciavald:
The button is misplaced, and I think it should be replaced by a bottom `close` button on the left of the others.
screenshot_20171206_165248

pciavald:
otherwise, please fix conflicting files.

anandsinghparihar:
SignAddressModal button and fields are same as QT wallet. clear all is basically for clear input. I've added close button at the top so a user can use it. :d

anandsinghparihar:
@pciavald please review PR

src/app/wallet/wallet/shared/signature-address-modal/signature-address-modal.component.ts
```diff
@@ -1,6 +1,6 @@
 import { Component, ElementRef, OnInit, ViewChild } from '@angular/core';
 import { FormBuilder, FormGroup, Validators } from '@angular/forms';
-import { MatDialog } from '@angular/material';
+import {MatDialog, MatDialogRef} from '@angular/material';
```
pciavald: this should contain spaces

src/app/wallet/wallet/shared/signature-address-modal/signature-address-modal.component.ts
```diff
@@ -39,7 +39,8 @@ export class SignatureAddressModalComponent implements OnInit {
     private _rpc: RpcService,
     private flashNotification: SnackbarService,
     private formBuilder: FormBuilder,
-    private _modals: ModalsService
+    private _modals: ModalsService,
+    private dialogRef: MatDialogRef<SignatureAddressModalComponent>
   ) {
   }
```
pciavald: closed

pciavald:
material effect is broken as it was at the beginning due to the html structure

screenshot_20171220_031530

anandsinghparihar:
@pciavald all issue fixed please check

allienworks:
Is the close button seriously manually added to each smaller modal? That's absurd, isn't it?
Why are we rewriting default material/angular functionality anyways? Can anyone please explain?

If it _must_ be done, I'd suggest doing that somewhere like in shared `modals.component` for large modals. These however are dialogs if I understand it correctly. So each should be used properly wherever fit.

allienworks:
So just to summarize our chat before:

- I don't think our current approach is good. Using "big" modals somewhere and using "smaller" somewhere else is OK of course. But there is no reason to have multiple code for all the modals, moreover adding _shared_ stuff (like close buttons) to each smaller modal separately.
- We should make **one** template for _all_ the modals! That means one layout for both small and big modals, so we can easily change all of them.
- We already have this in `modals.component` – all of these are using it now: create/restore wallet, lock/unlock wallet etc.
- All modals should be based on ^ `modals.component`
- We should have the ability to control each modal's behaviour by some parameters, examples:
  - size: big/small modal
  - close button: show/hide
  - etc..

Also, when I was going through changes in this PR (and all the modals), I noticed that we don't have ANY standardization of elements there. Titles are sometimes `h2`, sometimes `h3` or even `h4` (I've fixed those in my commit above). Same goes for "modal content" and "modal actions" (AKA buttons). This would be however solved by having the shared component, if I'm not mistaken.

This is obviously a big change for **1.2** (I won't block 1.1 release with this for now). We would have to remake all the modals in one, shared style. Regarding the styling, I would go with the current look of "small" dialogs (e.g. Address lookup) and ditch the ugly big modals (e.g. Unlock wallet). And yes, I'm aware that I'm the one who styled those big ugly ones in the first place :)

This would mean:

1. base all modals on shared `modals.component`, so they have the same functionality and looks
2. implement the "parameters" (show/hide close buttons, define widths etc.)
3. completely redesign the modals, because they make my eyes bleed (yes, that's one me ;)

Regarding this PR, I would probably leave it unmerged, before we can do it everything properly. If we decide this is a way to go, I (or some of you) can create a new PR with this modals' restructure.

@gerlofvanek @pciavald @anandsinghparihar – thoughts, comments, reasons anybody?

anandsinghparihar:
Okay, good suggestion but there is some reason why we used 'modals.component' for some component like all full width modal. (reusable old code with material modals).

You said
>Is the close button seriously manually added to each smaller modal? 

No after updating material modal close button is not working, that why close functionality added in every single modal. 

> implement the "parameters" (show/hide close buttons, define widths etc.)

Every time passing prams would not a good idea. 

Modal refactoring is really required it would be great if we have a small call with  @gerlofvanek @pciavald @allienworks   devs so it would batter to explaining things to each other? 

and for the moment should I remove some unusable close button because this PR is required on release.

allienworks:
> > Is the close button seriously manually added to each smaller modal?
>
> No after updating material modal close button is not working, that why close functionality added in every single modal.

Yes, it is manually added to each modal.. see this change: https://github.com/particl/partgui/pull/434/commits/c2497a5ade5a6e7af539dd695c529a860c2ab556

> >  implement the "parameters" (show/hide close buttons, define widths etc.)
>
> Every time passing prams would not a good idea.

Maybe, but it's still better idea then adding all the elements manually to each modal, IMO.

I think we can agree that we need a call to discuss and resolve this.

If this PR is needed to be merged, you can remove the close buttons as you write. However I'm not sure what will be left from this PR after you remove them (as this PR is just about adding close buttons..).

anandsinghparihar:
@allienworks after material update we dropdown and modal have a issue.. which is already reported an issue. 
so if the modal close class is not working then how I can close that modal without manual function. 

anyway, I am going to remove all close Icon.

pciavald:
Hi guys, 

I recommend that for 1.1 we just fix the closing buttons style (material effect and alignment), because there was a functionality problem before this PR that is fixed.

For 1.2 I recommend that we remove the big modals and replace them by views implemented by @kewde 's restructure PR. For instance, the big syncing modal is redundant with the sidebar, so first time you launch the wallet you get first a view of daemon download, then sync and create / import wallet prompt... and at the end of the process `open wallet` opens the app view that currently everything is in, in the form of modals to be removed.

I'll approve this PR once the styling issues are fixed, and i'll wait for your response before merging and creating an issue for 1.2.

allienworks:
> @pciavald I recommend that for 1.1 we just fix the closing buttons style (material effect and alignment), because there was a functionality problem before this PR that is fixed.

I already fixed the close button material-like shadows on focus. So we can probably merge this already. Someone please test it as well.

pciavald:
i've just checked and the shadow produces an ugly effect when clicking, screenshot ^ up there. Besides the button is too high by a few pixels which makes it look decentred.

allienworks:
> @pciavald i've just checked and the shadow produces an ugly effect when clicking, screenshot ^ up there. Besides the button is too high by a few pixels which makes it look decentred.

Fixed (if you're speaking about Sign/verify modal), can you please check & approve/merge?

PR #455, opened 2017-12-01
fix daemon restart

pciavald:
This PR fixes the daemon restart procedure and the `daemon-restart` RPC call needed by @anandsinghparihar. To do so, it implements a new `daemon.restart()` procedure.

It also fixes the logging : daemon output is cleaned out of double or missing newlines and the daemon output is disabled by default. Enable it by passing `-printtoconsole` after the `--`.

Enable the dev tools pane by passing `-opendevtools` after the `--`.


pciavald:
investigating a bug similar to https://github.com/atom/github/issues/1038

modules/daemon/daemon.js
```diff
@@ -24,27 +53,31 @@ exports.start = function(wallets, callback) {
...
     exports.check().then(() => {
       log.info('daemon already started');
       resolve(undefined);
 
     }).catch(() => {
 
       // TODO: only for some debug levels
-      // process.argv.push('-printtoconsole');
+      // TODO: pushed twice when restarting
+      process.argv.push('-printtoconsole');
```
kewde: Does this work again?
This was breaking my build, disabling resolved it.

anandsinghparihar:
First time its work fine, but when I deleted wallte.dat file and retest it then its return error `Unhandled promise rejection`. and sometimes empty modal open. 
please check screenshot, 
screenshot from 2017-12-06 22-07-43
screenshot from 2017-12-06 22-07-13

main.js
```diff
@@ -27,8 +27,10 @@ const daemon   = require('./modules/daemon/daemon');
 let mainWindow;
 let tray;
 let options;
+let openDevTools = false;
 
-let openDevTools = true;
+if (process.argv.includes('-opendevtools'))
```
kewde: :1st_place_medal:

modules/init.js
```diff
@@ -19,7 +19,7 @@ exports.start = function (mainWindow) {
     .then(()            => multiwallet.get())
```
kewde: https://github.com/particl/partgui/blob/dev/modules/multiwallet.js#L26

Does this work on windows?

kewde: ls -> linux?

pciavald: Hence the comment https://github.com/particl/partgui/blob/dev/modules/multiwallet.js#L22

Issue #484, opened 2017-12-07
Empty Dialog Modal 


anandsinghparihar:
Empty modal open if no wallet exists. display `Error: Loading block index.`

![empty-modal](https://user-images.githubusercontent.com/17725557/33711973-d6d5684c-db6b-11e7-8774-4151491948ea.png)

anandsinghparihar:
this error does not occur again but I saw this error sometimes we need to debug it in which case its display empty?

pciavald:
Can't handle this now, i have too much on the waiting line. @HenkSwardt ?

pciavald:
yes
i think it's a bit random
it depends how fast the daemon manages to start => if an RPC call is issued before the daemon is ready
this issue will be fixed anyway because the gui will be avoided from making calls while the daemon is not ready
but it doesn't mean that we shouldn't fix the empty modal problem
because it could pop for another reason

vikas-cis:
@pciavald I have added daemon error messages on the left side menu and get rid of the ugly modals. Please have a look and let me know 

Thanks

pciavald:
should be fixed

PR #487, opened 2017-12-08
get rid of daemon error handling modal and shows error message on the…

vikas-cis:
… left side menu

#484

### DIFF
--- a/src/app/core-ui/main/main-view.component.ts
+++ b/src/app/core-ui/main/main-view.component.ts
@@ -1,6 +1,7 @@
 import { Component, OnInit } from '@angular/core';
 import { Router, ActivatedRoute, NavigationEnd } from '@angular/router';
 import { Log } from 'ng2-logger';
+import { Observable } from 'rxjs/Observable';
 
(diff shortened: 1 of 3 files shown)


src/app/core-ui/main/main-view.component.ts
```diff
@@ -62,7 +62,10 @@ export class MainViewComponent implements OnInit {
     // Updates the error box in the sidenav whenever a stateCall returns an error.
     this._rpc.errorsStateCall.asObservable()
     .subscribe(status => this.daemonRunning = true,
```
kewde:
```
this.rpc.state.observe('blocks').throttle(val => Observable.interval(30000/*ms*/)).subscribe(block =>  this.rpc_update());
```
Use a throttle here

vikas-cis: yup, but not sure why you did polling for blocks as it refresh the whole table and I saw you removed my previous fix but that fix to prevent the whole table to refresh. 
@allienworks  we have discussion on this I guess ?

PR #510, opened 2017-12-14
Add "Send all" button/icon to Send page 

anandsinghparihar:
Add "Send all" button/icon to Send page


allienworks:
GFX LGTM, works nicely even when switching between Public/Private balances.

Haven't reviewed the TS code though.

anandsinghparihar:
this PR is cover: https://github.com/particl/partgui/pull/510

allienworks:
Code got a bit broken (probably due to my Send page redesign) – the "send all" button now dynamically don't change according to which balance is selected. Please fix.

> If you have 100 public coins, only 40 blind, select the public balance and then "send all", it shows correctly 100. But if you then switch to blind, the amount to send is still 100.

Other than that: OK by me.

allienworks:
Works perfectly now, thanks!
After someone approves TS code changes, we can merge.

PR #357, opened 2017-11-14
Narration implementation

allienworks:
- added status icons notifying about included narration
- narration content displayed in TX detail
- fixes https://github.com/particl/partgui/issues/350
- **still needs implementation of displaying narration content etc.** (I've just prepared the styling)

Preview:

## PRG TODO

- [x] show/hide narration icons in TX "headers" based on if narration's included or not
- [x] the same for narration content in TX detail
- [x] display narration instead of `>> Narration content here <<` + remove `TO_DO` class

PR #520, opened 2017-12-18
Display narration text

anandsinghparihar:
Narrations were already done and working, merged to develop (#357). Now on dev (and even on restructure), the narrations aren't working anymore.

I've checked if the changes from #357 are included and yes, they are there. Probably some logic changed during restructure and that might broke it.

So we need the narrations to display again.

### DIFF
--- a/src/app/wallet/wallet/shared/transaction.model.ts
+++ b/src/app/wallet/wallet/shared/transaction.model.ts
@@ -155,13 +155,12 @@ export class Transaction {
 
   /* Narration */
   public getNarration() {
-    if (this.n0) {
-      return this.n0;
-    } else if (this.n1) {
-      return this.n1;
-    } else {
-      return false
+    for (const key in this.outputs) {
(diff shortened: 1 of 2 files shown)


allienworks:
Works nicely! I added one small thing – display narration directly in the tooltip when hovering over narration icon:



Can someone pls review the code & merge?

src/app/wallet/wallet/shared/transaction.model.ts
```diff
@@ -155,13 +155,12 @@ export class Transaction {
...
   public getNarration() {
-    if (this.n0) {
-      return this.n0;
-    } else if (this.n1) {
-      return this.n1;
-    } else {
-      return false
+    for (const key in this.outputs) {
+      if (this.outputs[key] && this.outputs[key].narration) {
+        return this.outputs[key].narration;
```
pciavald: this will include only the first narration.
i guess it is good enough for this release

PR #528, opened 2017-12-19
added password verification

EagleEye25:
*added password verification on create wallet, to re-enter password that has been chosen for verification
*added notification messages on step 3,4 to save passphrase
*adjuested final message once create wallet process is complete


src/app/modals/createwallet/createwallet.component.html
```diff
@@ -223,7 +233,8 @@
 
       <div class="title">Congratulations!</div>
       <p>
-        Your new wallet was created successfully and set as the default wallet!
+        Your new wallet was created successfully and set as the default wallet!<br>
+        <b>Please ensure that your Passphrase is written down and stored in a sequre location!</b>
```
pciavald: secure *

src/app/modals/createwallet/createwallet.component.ts
```diff
@@ -112,16 +119,20 @@ export class CreateWalletComponent {
         if (this.isRestore) {
           this.step = 4;
         }
+        this.password = undefined;
+        this.passwordVerify = undefined;
         break;
       case 3:
         this._passphraseService.generateMnemonic(this.mnemonicCallback.bind(this), this.password);
+        this.flashNotification.open('Please remember to write down your recovery-passphrase', 'warning');
```
pciavald: remove hyphen in recovery-passphrase

src/app/modals/createwallet/createwallet.component.ts
```diff
@@ -112,16 +119,20 @@ export class CreateWalletComponent {
...
         break;
       case 4:
         while (this.words.reduce((prev, curr) => prev + +(curr === ''), 0) < 5) {
           const k = Math.floor(Math.random() * 23);
 
           this.words[k] = '';
         }
+        this.flashNotification.open('Please remember to write down your recovery-passphrase', 'warning');
```
pciavald: remove hyphen in recovery-passphrase

src/app/modals/createwallet/createwallet.component.ts
```diff
@@ -206,17 +217,34 @@ export class CreateWalletComponent {
...
+  /** verify if passwords match */
+  verifyPasswords() {
+    if (!this.validating) {
+      return;
+    }
+
+    if (this.password !== this.passwordVerify) {
+      this.flashNotification.open('Passwords Do Not Match!', 'warning');
+    } else {
+      // We should probably make this a function because it isn't reusing code??
```
pciavald: can you explain ?

pciavald:
- only one checkbox to show both password would be best for the recovery password verification, what do you think @allienworks ? Also, tab shouldn't stick to it before going to the next field.
- notification for step 4 should not be the same message than 3, i've adapted.

allienworks:
Could we rebase this to latest `dev`'s changes (= could someone resolve the conflicting files)? It's a bit too far behind; I'd look into it then..

EagleEye25:
the password component doesn't have the ability to read the checkbox from another component.. We would need to either add an output to get the value from the other component, or add a new checkbox on the creatwallet modal that passes the value into both the password components.
Please advise on which way we should proceed?

pciavald:
@EagleEye25 this will need refactor 
Can you please log an issue for 1.2.0 to refactor this to allow only one checkbox, and about my comment on the ^ code ^ ? 
Once you did I'll merge this

EagleEye25:
is this sufficient https://github.com/particl/partgui/issues/595?

Issue #344, opened 2017-11-14
redefine logging levels


pciavald:
We should have precise logging levels for the backend, that also logs the front. It will help debug the users' issues.

https://www.npmjs.com/package/electron-log

HenkSwardt:
Hi @pciavald 

Please confirm that this is now in place following the merge of #297 

Thanks

allienworks:
This has been solved already, right @pciavald?

pciavald:
sorry, nope it's not solved yet :)

pciavald:
moving to PR

PR #554, opened 2017-12-27
redefine logging levels

pciavald:
Following #344 logging levels (info, debug, error...) have been redefined.

Instead of using `-printtoconsole` and `-debug`, you can now simply use `-v` and `-vv`

### DIFF
--- a/main.js
+++ b/main.js
@@ -9,11 +9,29 @@ const platform      = require('os').platform();
 const rxIpc         = require('rx-ipc-electron/lib/main').default;
 const Observable    = require('rxjs/Observable').Observable;
 const log           = require('electron-log');
-
-log.transports.file.appName = (process.platform == 'linux' ? '.particl' : 'Particl');
-log.transports.file.file = log.transports.file
+/*
+ * Logger config
+ */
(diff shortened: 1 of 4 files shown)


main.js
```diff
@@ -9,11 +9,29 @@ const platform      = require('os').platform();
...
+log.transports.file.file     = log.transports.file
    .findLogPath(log.transports.file.appName)
    .replace('log.log', 'particl.log');
+
+log.transports.console.level = 'info';
+log.transports.file.level = 'debug';
+
+if (process.argv.includes('-v')) {
+  log.transports.console.level = 'debug'; /* most messages */
+  process.argv.push('-printtoconsole');   /* daemon output */
```
kewde: Can we maybe put the logging to console defined by another argument?
I'd like to be able to get general debug messages (Node backend), without the core logs.

kewde:
Can be merged after fixing the conflicts.

main.js
pciavald: Absolutely, thanks for noticing. Here's what I propose : 
-v : gui[debug]
-vv : gui[debug] + daemon[normal]
-vvv : gui[silly] + daemon[debug]

pciavald: @kewde

PR #558, opened 2017-12-28
Refactoring components

anandsinghparihar:
- Removed unused component (which also not use in a future. )
- Set return types of all function
- if no return type then set void
- Removed unused includes file
- Removed unused injected service


kewde:
Looking good! :+1:

PR #559, opened 2017-12-28
modal tweaks (layout, icons etc.)

vikas-cis:
Fixed layout issues in modal titles, added icons and more small fixes & tweaks.


src/assets/theme.scss
```diff
@@ -443,3 +443,8 @@ input {
 .capitalize {
   text-transform: capitalize;
 }
+
+// ------ Modal text lable alignment ------ //
+.modal-text-align {
+  margin-top: 20px !important;
```
allienworks: This doesn't belong into `theme.scss`! Keep this file clean. If you want to add something that is shared across modals, there's a `app/modals/shared/_modal-layout.scss` instead.

vikas-cis: actually this class is using in two components so I thought we can make it in once place

vikas-cis: I got it (y)

allienworks:
@vikas-cis also please add some descriptive description next time (for both PRs and issues)

src/assets/theme.scss
allienworks: Yes, that's why `_modal-layout.scss` is included in all modals.

vikas-cis:
@allienworks I have attached screenshot, yeah may be I miss to add some lines in it 👍

PR #585, opened 2018-01-04
Change History Tab labels into past tense

anandsinghparihar:


allienworks:
Wouldn't be "Balance Transfers" better? I don't think that everything has to be in past tense, here it doesn't make that much sense IMO.. What do you think?

HenkSwardt:
Im happy with that yes

Issue #587, opened 2018-01-04
Wallet auto-lock delay increased to 5 mins


HenkSwardt:
When the wallet is unlocked we automatically lock it after 30 seconds. This is a security feature to ensure your wallet is not left unlocked and unattended. Please change this delay time to 5 mins (300 sec).

dasource:
Please also add a notification on auto lock so the user knows. (bottom left)

HenkSwardt:
Sorry, forgot to add the above, thanks @dasource 

Please include a message as follow after the wallet has locked:
"Wallet auto locked"

vikas-cis:
@HenkSwardt do we need timer to show like "Wallet auto locked 4:59 so on" ?

Thanks

anandsinghparihar:
PR: https://github.com/particl/partgui/pull/589

PR #589, opened 2018-01-04
Wallet auto-lock delay increased to 5 mins

vikas-cis:
#587 
  When the wallet is unlocked we automatically lock it after 30 seconds. This is a security feature to ensure your wallet is not left unlocked and unattended. Please change this delay time to 5 mins (300 sec).


vikas-cis:
@pciavald i have added timer, can you please have a look 

Thanks

vikas-cis:
@allienworks Can you please check if we need graphical things for the message ?

allienworks:
@vikas-cis may I have one more request? The countdown is now visible even a bit after it reaches 0 (looks like something like `-5:00` flashes there for a fraction of second). Could you hide the countdown ~1 second earlier so this won't happen?

allienworks:
Countdown styled, all known issues fixed. It's neat! :)

PR #596, opened 2018-01-05
fix wallet "tab" being displayed twice

EagleEye25:
made it so that if the wallet.dat isnt encrypted that the gui will only show the add account button on the sidebar

### DIFF
--- a/src/app/core-ui/main/main-view.component.html
+++ b/src/app/core-ui/main/main-view.component.html
@@ -20,7 +20,7 @@
       </mat-list><!-- .sidebar-item-group -->
     </mat-content><!-- .sidebar-section -->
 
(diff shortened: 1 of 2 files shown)


pciavald:
very strange, on this branch and even after merging dev in, i have this GFX : 
screenshot_20180105_134105
as opposed to the normal : 
screenshot_20180105_134241

pciavald:
can you check what happens when you start creating a wallet, quit app at the "copy your words" step and launch again ?

pciavald:
i confirm this issue.

screenshot_20180108_134329

HenkSwardt:
Could this and #614 have something in common?

allienworks:
Truly weird behaviour. For some reason another `box-shadow` was added there, not sure why. I've fixed that.

Issue #560, opened 2017-12-29
Pagination per page not update while we search (All pages search option Receive/Addressbook)


vikas-cis:
How to Reproduce:
1)When you are on address book page, Initially you are on first page and you do the search and get the results correctly.
2)But when you move to second page and try to search result for the same page or previous page result it will not give you any result.

![screen shot 2017-12-29 at 4 03 18 pm](https://user-images.githubusercontent.com/20254889/34435226-f65ec06c-ecb1-11e7-81d9-2e3994e57826.png)
![screen shot 2017-12-29 at 4 03 29 pm](https://user-images.githubusercontent.com/20254889/34435227-f691b684-ecb1-11e7-828e-3ad904a7d7da.png)

anandsinghparihar:
PR: https://github.com/particl/partgui/pull/561

PR #607, opened 2018-01-08
added filter options on addressBook

EagleEye25:
*added the ability to filter address book by using filters
*updated reg expressions in addressHelper utility

### DIFF
--- a/src/app/wallet/shared/util/utils.ts
+++ b/src/app/wallet/shared/util/utils.ts
@@ -147,9 +147,9 @@ export class Duration {
   }
 
 export class AddressHelper {
-  addressPublicRegex: RegExp = /^[pPrR][a-km-zA-HJ-NP-Z1-9]{25,35}$/;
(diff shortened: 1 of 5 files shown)


pciavald:
@HenkSwardt this is @allienworks 's job if i'm correct, will review when styled

allienworks:
Styled.

**However there is still one issue** – when selecting some address type from the filter, the pagination doesn't get updated (number of results & pages stays the same, e.g. even if 80% addresses are missing).

I believe we had this issue somewhere already, maybe in Address lookup modal, so you might look there for inspiration.

HenkSwardt:
@EagleEye25 , Please look at above point from @allienworks 

Thanks

kewde:
#560 example of the pagination updating

pciavald:
@EagleEye25 @HenkSwardt is that ready for review ?

EagleEye25:
@pciavald yes it it ready for review i believe

allienworks:
Tested and works nicely. Needs code review.

pciavald:
@EagleEye25 that works, please use this.log.{e,d} instead of console.log :)

PR #609, opened 2018-01-08
Feature/zap

pciavald:
Ability to ZAP the remaining balance not in cold staking yet.


pciavald:
@gerlofvanek @allienworks what needs to be done here : 
- style zap button (I was thinking of a lightning logo)
- dialog to tell more about the zap functionality and inform the user of transaction cost, confirm or cancel

pciavald:
*note to self*

https://www.zerobin.net/?bdf999af53855149#TAVaPp5xWrEQdE8ioM18ubBgLaeaWZOC2uSfC5rZgHA=
https://www.zerobin.net/?acf69f9dcd75ac6f#8OS8VeWpaa+AVKXYD3QVnXA46Bsqpl5tREhzBQenSEY=
https://www.zerobin.net/?e992ecbe67b48b77#0gzqMp3li4I5DcKwP8HtmjIpI63GjVR86CgK4BTbf0A=

pciavald:
- listunspent before zap to avoid double zap
- revert cold staking
- what happens when I spend after cold staking ? staking may be completely stopped for the maturation time

allienworks:
@anandsinghparihar @vikas-cis can any of you please implement these changes please?

- "zap" button in Cold Staking widget will now open small modal
- inside this modal will be the new "zap" button, which will have the "zapping" functionality (as currently has the one in the widget)
- add the "cancel" button to the modal as well (which will close the modal)

I'll add the rest of the stuff then, style it etc.

Btw @pciavald good thinking about TX cost. We don't have that yet, in the GUI, _at all!_ This needs to be done as well.

anandsinghparihar:
@allienworks I am on it

anandsinghparihar:
@allienworks please check

src/app/wallet/overview/widgets/coldstake/coldstake.component.ts
```diff
@@ -60,17 +70,94 @@ export class ColdstakeComponent implements OnInit {
...
+                  amount: sum_inputs,
+                  script: script
+                }], '', '', 4, 64, true, JSON.stringify({
+                  inputs: inputs
+                })]).subscribe(info => {
+
+                  this.log.d('zap sendtypeto simulate', info);
+
+                  // TODO: ask user to confirm info.fee in a modal
+                  this.dialog.open(ZapWalletsettingsComponent);
```
pciavald: @anandsinghparihar can you pass the `info.fee` to the dialog here please ?

src/app/wallet/overview/widgets/coldstake/zap-walletsettings/zap-walletsettings.component.ts
```diff
@@ -0,0 +1,36 @@
...
+@Component({
+  selector: 'app-zap-walletsettings',
+  templateUrl: './zap-walletsettings.component.html',
+  styleUrls: ['./zap-walletsettings.component.scss']
+})
+export class ZapWalletsettingsComponent {
+
+  private log: any = Log.create('zap-walletsettings');
+
+  fee: number = 42;
```
pciavald: in order to assign it here

anandsinghparihar:
@pciavald request changes pushed !!

pciavald:
# warning

at this stage, clicking `UNDO` will unset your pkey, don't do it
at this stage, ZAP and UNDO buttons display how many transactions will be processed by the action, buttons with `0` will not be displayed later
for instance, if you have `ZAP 0` it means you don't have anything to zap ; you can still open the modal (won't be allowed after) and the transaction fee cannot be computed as the transaction is not valid.

allienworks:
Styled. I haven't tested the functionality though. Ready for code review.

src/app/wallet/overview/widgets/coldstake/revert-coldstaking/revert-coldstaking.component.ts
```diff
@@ -0,0 +1,89 @@
...
+    this.log.d('undo coldstaking');
+
+    this._rpc.call('walletsettings', ['changeaddress', {}]).subscribe(res => {
+
+      this.log.d('disable coldstaking', res);
+      if (res.changeaddress !== 'cleared') {
+        return false;
+      }
+
+      // TODO: update status of cold staking widget
```
pciavald: @anandsinghparihar can you update the cold staking widget state here ? 
Currently, when reverting cold staking, key is removed but widget is not updated so it seems cold staking is still activated, while it is not. What we need is that once we reverted, cold staking widget returns to the state where there's a "+" sign to add a pkey.
peek 2018-01-10 21-13

pciavald:
I still need to address that : 

[dasource] particl-cli extkey account | jq .chains[0].num_derives Check to make sure that this is the "active_external" chain and +1 the num_derives

And make sure that a stealth address is present on a blank wallet or create one with `getnewstealthaddress`

anandsinghparihar:
@pciavald , update cold staking widget after clicking on revert.

src/app/wallet/overview/widgets/coldstake/revert-coldstaking/revert-coldstaking.component.html
```diff
@@ -0,0 +1,30 @@
+<h2 mat-dialog-title>Revert Cold Staking</h2>
```
dasource: Disable Cold Staking

pciavald:
@HenkSwardt with no further input from dasource or tecnovert and the above test ^ being fixed, that task is done.

src/app/wallet/overview/widgets/coldstake/revert-coldstaking/revert-coldstaking.component.ts
pciavald: done elsewhere

anandsinghparihar:
@pciavald  I am fixing `ng-test` and `ng-lint`

allienworks:
I've tried this on mainnet, wanted to zap the few remaining %.. Zapped two times, two TX went away, but the progress haven't moved at all.

pciavald:
@allienworks zapped two times ? shouldn't be possible :/

allienworks:
@pciavald I've zapped once first, a balance transfer TX appeared, but the progress % stayed the same, so I waited a few blocks and when still nothing happened, I've zapped for the second time. I was running this branch via `npm run start:electron:dev` (without the `-testnet`), not sure if that's the right way to run this on mainnet.. I did that 14/1 22h and left the wallet online unlocked for staking since, hoping it will come through. About another 3% went to cold staking since, but I still have a few % remaining.

allienworks:
@pciavald after our session, my _testnet_ wallet (which is not even encrypted) doesn't display the CS widget:



I know it's not a standard behaviour (having your wallet *un*encrypted), but this should be solved anyway – some of the users might be affected by this (at least the newcomers indeed). I propose having the same style of the widget as "unlock to check state" but with a message like "encrypt wallet to setup cold staking" or something like that (with button pointing to wallet encryption modal ofc).

anandsinghparihar:
@pciavald test cases is fixed

pciavald:
this still has a bug, @allienworks reported wrong zapping logic when zapping on several utxos. We need to investigate this.

Issue #610, opened 2018-01-08
Add All serach Input with a clear button


anandsinghparihar:
Right now all search input doesn't have a clear button, we need to clear that input by backspace. so I hope we should add clear button.
![input angular material](https://user-images.githubusercontent.com/17725557/34671979-b7802dce-f4a2-11e7-8106-b2a4ffe99e38.png)

anandsinghparihar:
PR: https://github.com/particl/partgui/pull/611

PR #611, opened 2018-01-08
Input clear button

anandsinghparihar:
Add All search

 Input with a clear button #610 
>> Right now all search input doesn't have a clear button, we need to clear that input by backspace. so I hope we should add clear button.

here is a preview of a clear button : 
input clear button

### DIFF
--- a/src/app/wallet/wallet/address-book/address-book.component.html
+++ b/src/app/wallet/wallet/address-book/address-book.component.html
@@ -12,7 +12,7 @@
                placeholder="Search Addresses" (keyup.escape)="query = ''">
 
         <mat-icon *ngIf="query" (click)="query=''" matTooltip="Clear" fontSet="partIcon"
-                  fontIcon="part-cross"></mat-icon>
+                  fontIcon="part-clear-all"></mat-icon>
 
         <mat-icon *ngIf="!query" fontSet="partIcon" fontIcon="part-search" fxHide.lt-md matTooltip="Search For Address"
                   matTooltipPosition="before"></mat-icon>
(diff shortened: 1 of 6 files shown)


allienworks:
I'm honestly not that sure if this is absolutely needed, but OK. Styled.
IMO there are more urgent tasks at hand, so please try to do those first before making up other (not necessary) tasks.

anandsinghparihar:
@allienworks that time I don't have any task ... anyway thanks for your suggestion  :d

allienworks:
@anandsinghparihar I get that. Have a look at the current milestone when you're out of work. There's a lot of stuff needed to be done for the next release (always).

anandsinghparihar:
@allienworks I don't want  argue ask to Henk :)

pciavald:
@anandsinghparihar @allienworks I edited the thread conversations, these matters are better discussed on Matrix 👍 cheers :)

anandsinghparihar:
@pciavald 
> can you please activate escape to clear & enter like in history search
Enter is not required because of all search fire on input changes. I've added escape for clear search field. 

@allienworks now you can work on the design part.

allienworks:
Restyled.

This is how icons are organized now:

- `part-circle-remove` as "close/delete" (for closing modals, deleting entries)
- `part-clear-all` as "clear" (e.g. here for clearing text from inputs or whole forms)
- `part-cross` as "cancel"

I don't think we should unify all those 3 into 1; after all, there are differences between them. Those icons that aren't fixed according to this ^ are fixed already in #657 (FYI).

Issue #595, opened 2018-01-05
Refactoring password component to allow ony one "view password" checkbox


EagleEye25:
# Describe bug/issue
*Currently have two checkboxes for verification of passwords.
*https://github.com/particl/partgui/pull/528#issuecomment-355380296, will have to be refactored in v1.2, to only use on checkbox, to hide and display passwords

vikas-cis:
#617

PR #617, opened 2018-01-08
refactor passwordcomponent to show password for both text field using…

vikas-cis:
… one checkbox

#595


pciavald:
Please disable tab focus on the checkbox.
- click field 1
- tab
- should be on field 2, but is on checkbox

Issue #620, opened 2018-01-08
cold staking widget state


pciavald:
# unlock cold staking widget

### case

- start wallet
- unlock **cold staking widget**
- wallet is **fully unlocked**
- widget shows **paused**

![peek 2018-01-08 22-08](https://user-images.githubusercontent.com/3384097/34692526-ca5f6df6-f4c0-11e7-9bfa-01cf4d79b44b.gif)

### expected behaviour

- start wallet
- unlock **cold staking widget**
- wallet is **unlocked for staking**
- widget shows **Activation in progress** or **Activated** if at 100%

# full unlock

### case

- start wallet
- **fully unlock** wallet
- widget shows **paused**

![peek 2018-01-08 22-09](https://user-images.githubusercontent.com/3384097/34692538-d390a0a2-f4c0-11e7-97ee-b14cd4981c40.gif)

### expected behaviour

- start wallet
- **fully unlock** wallet
- widget shows **Activation in progress** or **Activated** if at 100%

# staking only

### case

- start wallet
- unlock wallet **for staking only**
- widget is still locked

![peek 2018-01-08 22-10](https://user-images.githubusercontent.com/3384097/34692554-dbb35504-f4c0-11e7-8778-6d4df7281329.gif)

### expected behaviour 

- start wallet
- unlock wallet **for staking only**
- widget shows **Activation in progress** or **Activated** if at 100%

# 100%

### case

- start wallet
- unlock **cold staking widget**
- zap
- balance is at 100% 
- lock wallet
- widget shows **Paused**

![screenshot_20180108_221241](https://user-images.githubusercontent.com/3384097/34692727-61f01d0a-f4c1-11e7-899c-268181bf05f7.png)

### expected behaviour

- start wallet
- unlock **cold staking widget**
- zap
- balance is at 100% 
- widget shows **Activated**

### case

- start wallet
- unlock **cold staking widget**
- zap
- balance is at 100% 
- widget shows **Activation in progress**

![screenshot_20180108_221302](https://user-images.githubusercontent.com/3384097/34692733-68b8fa1c-f4c1-11e7-861d-e5ff54fcf484.png)

### expected behaviour

- start wallet
- unlock **cold staking widget**
- zap
- balance is at 100% 
- widget shows **Activated**

pciavald:
@HenkSwardt

anandsinghparihar:
After entering the cold staking public Key  empty modal is open, So I guess that modal should be close 
[...]

PR #624, opened 2018-01-09
cold staking widget state different scenarios fixing

vikas-cis:
#620

### DIFF
--- a/src/app/wallet/overview/widgets/coldstake/coldstake.component.ts
+++ b/src/app/wallet/overview/widgets/coldstake/coldstake.component.ts
@@ -18,7 +18,7 @@ export class ColdstakeComponent implements OnInit {
   coldStakingEnabled: boolean = undefined;
   stakingTowardsCold: boolean = undefined;
   activation: string = 'Activation in progress';
-
+  public encryptionStatus: string = 'Locked';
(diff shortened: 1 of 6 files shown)


src/app/wallet/overview/widgets/coldstake/coldstake.component.ts
```diff
@@ -64,6 +65,9 @@ export class ColdstakeComponent implements OnInit {
     if (this.coldstakeProgress < 100) {
       setTimeout(this.rpc_progressLoop.bind(this), 1000);
     }
+    if (this.coldstakeProgress === 100) {
+      this.activation = 'Activated';
```
pciavald: we still need to poll here in case there's a new transaction after full activation

pciavald:
### unlock widget does not have the right behaviour
Wallet should be **unlocked for staking only** when unlocking the widget, instead of the current **fully unlocked** with timer.

### unlock wallet for staking only does not have the right behaviour
When unlocking for staking only, coldstaking is at 0%. Locking and fully unlocking the wallet updates the coldstaking at 100% (real here).
peek 2018-01-09 23-34

pciavald:
Perfect, @vikas-cis .

Issue #633, opened 2018-01-10
Notification issue with fresh wallet and found timer showing negative value


vikas-cis:
How to reproduce Notification issue
1)Fresh wallet
2)Send some parts to the fresh wallet address
3)You will notified with a notification 
4)When you moved history to receive or vice-versa you will get notification again and again.

Unlocking timer : Shows negative value sometimes while unlocking the wallet.

vikas-cis:
#634

Issue #627, opened 2018-01-10
1 Sat balance in Blind to display 0 Balance


HenkSwardt:
When you select to transfer all from blind to public, the transaction needs to leave 1 sat behind. This now case the blinded balance on SEND page to display as below:

![image](https://user-images.githubusercontent.com/28799964/34736615-406d6e38-f57c-11e7-8555-4fc69ff01fe5.png)

Please display "0" balance in the case of the above scenario.

HenkSwardt:
Based on group discussion we had a look at the following:

Because of the way the blinded txns work you can't transfer all to normal part, it needs to leave a change output of at least 1 sat behind. So for now if balance < 0.000001 then we will display "0" balance

kewde:
Please note that it accumulates, so 2 seperate txns that transfer all coins will leave 2 sats behind (total, because they are 1 sat each).

HenkSwardt:
Following further discussion we have decided to implement the following for this scenario:

If the blind_balance < 1000 sats display a "?" next to blind balance which on hover says the following as a tooltip:

"It is normal to have a very small balance in Blind even after transferring out everything. This is due to the way CT works and part of the privacy feature."

PR #634, opened 2018-01-10
added question mark with tooltip for blind balance and also done some…

vikas-cis:
… tweaks related to transaction service and unlocking timer

#633  #627


vikas-cis:
@allienworks Can you please check if we required to do some tweaks for 'question mark', the issue link is here #627

allienworks:
Functionality works for me. Still needs code review.

PR #656, opened 2018-01-11
debug console wrong command notification  

anandsinghparihar:
>currently, debug console refuses to input a wrong command, with no feedback to the user. Insert a flash notification to inform that a wrong command was issued.


pciavald:
Ah ! thank you, that was much needed.

PR #664, opened 2018-01-12
Small changes for zap(blance transfer) tx record

anandsinghparihar:
after zapping and cancel zap, I see history list and overview page having an error, Transaction modal having some issue.  because zap balance transfer does not return output (address etc.. )


pciavald:
already done here. 
https://github.com/particl/partgui/pull/609/files#diff-e472f9e0c2b0351d9d8e9e3e964de3e9L36

pciavald:
won't merge zap right now so including this into dev

Issue #681, opened 2018-01-16
mac wallet creation not working


pciavald:
create wallet fully
wild unclosable blank modal appears
close wallet
open wallet 
go to receive page 

<img width="1263" alt="capture d ecran 2018-01-16 a 20 03 30" src="https://user-images.githubusercontent.com/3384097/35007025-95c34dcc-faf8-11e7-9425-b29441ed4dc2.png">

HenkSwardt:
@vikas-cis tested the creation of Mac wallet and confirmed that Daemon still not connecting.

pciavald:
@HenkSwardt 
1. is solved in #690 
2. I will do a PR to fix

pciavald:
2. fixed in #691

PR #689, opened 2018-01-17
Mac daemon issue

vikas-cis:
#681


pciavald:
based on zap, please base on dev

PR #725, opened 2018-01-19
app title

pciavald:
fixing app names after name change


kewde:
Already in the commit from TV in release/1.1.0-RC2

PR #726, opened 2018-01-19
fix error on creation screen of no pass entered

EagleEye25:


pciavald:
@EagleEye25 sis you just invert the conditions ?

Issue #710, opened 2018-01-18
Ctrl+V is not working on mac


vikas-cis:
How to Reproduce

1)Initialize the wallet
2)Create wallet 
3)Copy all the mnemonics
4)Try to fill vacants mnemonics using Ctrl+V
I have tested in Particl 1.1.0-RC1 (testnet) (internal)
I have also created build from dev and tested, Having same issue 

Thanks

pciavald:
on mac it's command-V

pciavald:
I tried to paste a word in the vacant mnemonics on mac using `command+v` and I confirm it does not work.

vikas-cis:
#740

PR #740, opened 2018-01-22
ctrl+v issue fix for OSX

vikas-cis:
#710


pciavald:
@vikas-cis does this work for command+V ? ctrl+V is not standard at all on OSX.

vikas-cis:
@pciavald it will work for both 

thanks

pciavald:
that pastes twice on linux

vikas-cis:
thanks @anandsinghparihar  I just want multiple people can test on different machine so we have clear look of the functionality that implemented works on different machine  👍

pciavald:
@vikas-cis can you try to make @coveralls happy ? I'll test again

pciavald:
works on linux and windows. Also works on Mac with both cmd+V and ctrl+V, which i think we do not want. Ctrl+V is not a way to paste on mac, only cmd+V should work.

vikas-cis:
@pciavald I was just wondering if user change the setting from preference ?

Thanks

pciavald:
@vikas-cis on mac you can change the order of keys (command, ctrl, alt) so that command is triggered when you push the ctrl key, but the actual paste at the OS level is with command+V even though you can configure to press ctrl, afaik.

kewde:
To be honest, I don't think we should allow CTRL + V on _one_ component and not on the others.
Either we do this for the whole application, or not at all.

pciavald:
@kewde before merging we should test it on win, mac & linux

Issue #753, opened 2018-01-23
broken layout


pciavald:
observed on mac build

![screen shot 2018-01-23 at 2 19 53 pm](https://user-images.githubusercontent.com/3384097/35288730-cf16e3be-0065-11e8-9f3c-bda93ccf3301.png)
![screen shot 2018-01-23 at 5 47 09 pm](https://user-images.githubusercontent.com/3384097/35288751-d9c84384-0065-11e8-919a-e2a651a30c1d.png)
![screen shot 2018-01-23 at 5 47 58 pm](https://user-images.githubusercontent.com/3384097/35288757-dde6eeac-0065-11e8-8801-b043615c36eb.png)
![screen shot 2018-01-23 at 5 46 58 pm](https://user-images.githubusercontent.com/3384097/35288791-eed0e146-0065-11e8-8abd-48af478b2f91.png)

vikas-cis:
@pciavald at least you have also got this issue 👍

HenkSwardt:
This was reported to be broken on Linux build as well please see #694

thecarterzim:
Was coming to add this bug, but see it already exists! Thanks.

HenkSwardt:
This has been logged as a duplicate. Closing issue in favor of #694

allienworks:
@pciavald what's the current state of this? Are you having still the same problems as with the screens above? As mentioned in https://github.com/particl/particl-desktop/pull/758#issuecomment-360430998 the Electron 1.8.2 should fix this, so all we can do is wait for the release.

HenkSwardt:
The upgrade to 1.8.2 did not resolve this issue as expected. Further investigation has been done and it seems that a change to muon will help with resolving this. 

Please see - https://github.com/particl/particl-desktop/pull/807

This needs to be reviewed and tested.

pciavald:
fixed in #807

vomm:
I´ve got few times a little displacement....like one time in 10-20 clicks on Staking node. Windows 10 - 64bit

![rc4_2](https://user-images.githubusercontent.com/28147511/36157437-1319fac2-10da-11e8-8d12-0aab22390185.jpg)

imybb:
observed on Windows RC4 Sandboxed version 

![windows_rc4_out_of_alignment](https://user-images.githubusercontent.com/30150617/36157527-e3e2e4ec-10d1-11e8-976b-c7b93ec10c80.png)

vomm:
and another one.. :)
[...]

PR #758, opened 2018-01-24
Modal layout unification

### LINKED ISSUE #694: Modal design issue in Linux build
@allienworks I've created a build using latest code and I found following design issue, I guess Vikas is already reported that issue in Mac but the same issue is in Ubuntu as well. 
screenshot from 2018-01-17 21-39-58
screenshot from 2018-01-17 21-37-35
screenshot from 2018-01-17 21-36-45

AllienWorks:
WIP PR containing experiments and debug for the modals' layout.
Closes #694, #753 

- I've stripped down a lot (close buttons, styles etc.)
- there is a bug in former "big" modals - when closing, they somehow get smaller to a small square (that's because of the padding there) and this little square disappears then.. I'm not sure why other modals (the current smaller dialogs) don't behave like that.. can someone investigate?


allienworks:
Tried to disable animations (https://material.angular.io/guide/getting-started#step-2-animations), after that I can't reproduce the issue. Can @vikas-cis and @anandsinghparihar confirm?

vikas-cis:
@allienworks it seems the issue is not appearing but the modal looks ugly to me

allienworks:
So after reading a lot of different issues in angular/material and related, I found a lot of other devs ([1](https://github.com/angular/material2/issues/4227#issuecomment-345308783), [2](https://github.com/angular/material2/issues/8517), [3](https://github.com/angular/material2/issues/8588)) having the same issues as we have (moved elements or their parts).

Looks like it's a problem with Electron and not Angular Material, apparently after updating to latest [Electron (1.8.2) beta](https://github.com/electron/electron/releases/tag/v1.8.2-beta.3) everything works as expected ([source](https://github.com/angular/material2/issues/8588#issuecomment-353439856)).

Can we try that? At least in some testing branch (like this one)? Would be great to pin point the problem, so at least we know where it originates. We could focus on other issues and when Electron 1.8.2 is stable and out, simple update will hopefully do.

allienworks:
I've unified the modals look and code.

**The situation**
So, the [Material dialogs/modals](https://material.angular.io/components/dialog/overview) are built by 3 parts. The title (`mat-dialog-title`), content (`mat-dialog-content`) and optionally some action buttons (`mat-dialog-actions`). Currently, some of the modals (known as "small" dialogs, e.g. Add new address to Address book, Address lookup, Console, QR modal) are using exactly this setup. So that's good.

Then we have the "big" modals (e.g. Create wallet, Encrypt wallet, Cold staking), which are built with the `modals.component`. However in that component, we have just one "place" where we insert content (think `mat-dialog-content`). Therefore everything these modals are made of is inserted to "modal's content". It doesn't sound bad, but we can't have proper titles and/or the action buttons there – if you insert those in the content itself (and not where they belong – `mat-dialog-title` and `mat-dialog-actions`), it's breaking the layout. You can try to pull this PR and see it action in the Encrypt/Cold staking modals.

> This is all related to remake of modals (we have to do it either way). The issue with "jumping content" should be resolved by updating Electron to 1.8.2 as I wrote here earlier, just FYI.

**What needs to be done**
_IF_ we want to use templates (and I think that's surely a good idea; let's not repeat ourselves), we need to build one master template (think `modals.component`), _BUT_ we need to include all of the 3 parts there – so we can inject all types of content there (title, content and action buttons).

I'm not quite sure if it's possible, but it should be. Can someone look into it? I would start by adapting the `modals.component` (or you can create a new one, e.g. `dialogs.component`) and then implement it into `coldstake.component` (that one is easy enough).

allienworks:
@kewde absolutely. Let me know, we'll work it out.

allienworks:
@vikas-cis @anandsinghparihar can you test again, if the "jumping content" bug is still present? Thanks!

vikas-cis:
screen shot 2018-01-29 at 8 10 13 pm

@allienworks  it seems when I try to open cold staking modal from widget 4-5 times and then I try 6th time then it stuck not sure why ?

vikas-cis:
@allienworks check boxes for coldstaking widget and unlock modal are still moving as shared you via screenshare

Thanks

allienworks:
@vikas-cis the screen you shared is the result of all I wrote above. We need to rework the modals. Already on that with @kewde.

anandsinghparihar:
@allienworks please consider this  as well
cross icon issue

rynomster:
Are we still going to need this?

AllienWorks:
@rynomster yes, absolutely. We need to rework the modals altogether. Current setup isn't done properly (speaking about layout).

AllienWorks:
Current PR: https://github.com/particl/particl-desktop/pull/1025

PR #764, opened 2018-01-26
IPC refactor 

kewde:
Refactored the IPC service

- [x] you can now register listeners on the renderer side
- [x] added testcase/example for ZMQ integration
- [x] move code, streamline ipc a bit more
- [x] moved notifications


main.js
```diff
@@ -1,26 +1,27 @@
 
-const electron      = require('electron');
```
kewde: aww, my formatter killed this.. Can someone check this out?

kewde:
The karma tests are giving an error due to the fact that the IPC is not available in the Chromium browser.
We can look into running karma with electron however:
https://github.com/twolfson/karma-electron

Or I can make a small "mock" test to inject, which will emulate the IPC and get rid of the errors.

main.js
pciavald: what do you mean @kewde ?

pciavald:
@kewde won't merge this in dev right now, right ?

kewde:
@pciavald nope, it's for the 1.1.1

main.js
kewde: @pciavald I liked the previous formatting, where the '=' are all aligned vertically.

PR #769, opened 2018-01-29
Improve router

kewde:
Blocked by #764 

- [x] Reworked the router, we now have a core-router.
- [x] Move the big modals to the views


allienworks:
Styled the splash loading and cleaned-up a bit:

pciavald:
so many conflicting files 😮 
@allienworks animations now work while loading ?

allienworks:
> @pciavald animations now work while loading ?

Technically no, the loading screen itself is now empty and just flashes for a second. Then this "new" loading view takes place and it works. Maybe it's also because of the animations being done in pure CSS..

anandsinghparihar:
@kewde I 
- [ ] Wallet Initially with blank modal open (encryption modal I guess).
black modal box


- [ ] Zap functionality is not working not working.

- [ ] If I created wallet without encryption then after wallet created I cant able to set the encryption key.

anandsinghparihar:
If I created a wallet using enter button then it says wallet created but actually not !! 
https://www.useloom.com/share/1a29d7cf33fc4a8aaee9e980bc1995db

kewde:
state service failure is fixed in #797

Issue #684, opened 2018-01-16
mac dmg background not set


pciavald:
<img width="545" alt="capture d ecran 2018-01-16 a 20 11 28" src="https://user-images.githubusercontent.com/3384097/35008176-e652b5fe-fafb-11e7-89ea-6c4b8de8cd18.png">

vikas-cis:
@allienworks @gerlofvanek can you please tell me which image  we can use for background ?

gerlofvanek:
<img alt="" src="https://i.imgur.com/dzJFAq2.png">

vikas-cis:
@gerlofvanek Is the image is added ?

Thanks

vikas-cis:
@pciavald @gerlofvanek  Please have a look the screenshot this how it looks now after adding the background image in mac
![screen shot 2018-01-29 at 4 27 45 pm](https://user-images.githubusercontent.com/20254889/35507051-7ceb2fea-0511-11e8-8366-914fa3431f36.png)

PR #770, opened 2018-01-29
add background image for mac dmg

vikas-cis:
#684


kewde:
Oops, there is still some work on this it seems.

PR #778, opened 2018-01-31
Send All (1e-8) from Blind to Public gives error fix

### LINKED ISSUE #776: Send All (1e-8) from Blind to Public gives error
Please find the attached screenshot 
screen shot 2018-01-31 at 4 07 18 pm
screen shot 2018-01-31 at 4 07 31 pm
screen shot 2018-01-31 at 4 07 48 pm

vikas-cis:
#776

### DIFF
--- a/src/app/wallet/wallet/send/send.component.ts
+++ b/src/app/wallet/wallet/send/send.component.ts
@@ -70,6 +70,7 @@ export class SendComponent {
     this.send.input = 'balance';
     if (this.type === 'balanceTransfer') {
       this.send.toAddress = '';
+      this.send.output = 'blind_balance'
       this.verifyAddress();
     }
     this.updateAmount();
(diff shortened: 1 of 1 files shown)


src/app/wallet/wallet/send/send.component.ts
```diff
@@ -120,6 +121,11 @@ export class SendComponent {
       this.send.validAmount = false;
       return;
     }
+
+    if (this.send.amount === 1e-8) {
```
anandsinghparihar: @vikas-cis  can we set amount `0` if the amount equals to `1e-8` ?

vikas-cis: @anandsinghparihar its not `0` it is 1 SAT I guess which is a valid amount. So I think we don't need to show `0`

src/app/wallet/wallet/send/send.component.ts
```diff
@@ -70,6 +70,7 @@ export class SendComponent {
     this.send.input = 'balance';
     if (this.type === 'balanceTransfer') {
       this.send.toAddress = '';
+      this.send.output = 'blind_balance'
```
kewde: why this?

vikas-cis: @kewde yeah if you are on "send payment" and check the `blind` and then moved to "Balance Transfer" in that case `public` is checked on both side which is incorrect.

PR #789, opened 2018-02-03
Update check

kewde:


modules/init.js
```diff
@@ -10,20 +10,58 @@ const multiwallet   = require('./multiwallet');
...
+
+  // Done -> means we have a binary!
+  if(status === 'done') {
+    log.debug('daemonManager returned done, starting daemon!');
+    multiwallet.get()
     // TODO: activate for prompting wallet
     // .then(wallets       => ipc.promptWalletChoosing(wallets, mainWindow.webContents))
     .then(chosenWallets => daemon.start(chosenWallets, daemonStarted))
-    .catch(err          => log.error(err));
+    // .catch(err          => log.error(err));
```
pciavald: why removing that catch ?

modules/rpc/cookie.js
```diff
@@ -106,7 +106,6 @@ function getAuth(options) {
     auth = fs.readFileSync(COOKIE_FILE, 'utf8').trim();
   } else {
     auth = undefined;
-    log.debug('could not find cookie file! path:', COOKIE_FILE);
```
pciavald: we should let this, it's a good debug output that should not be displayed in release as logging levels will have been adjusted

kewde: I'll re-insert it, it was annoying me

Issue #743, opened 2018-01-22
Refactor state & rpc-state


kewde:
Multiple issues described here.

### turn RpcStateClass into RpcStateService
OnDestroy life cycle hooks only work in angular _services_
https://github.com/particl/particl-desktop/blob/a149e579cf512ba1d0bf50ff8cca3b6e6b406f15/src/app/core/rpc/rpc-state/rpc-state.class.ts#L5

### shareReplay in state
Do we really need to shareReplay? Wouldn't this cause it to cache forever? 
https://github.com/particl/particl-desktop/blob/a149e579cf512ba1d0bf50ff8cca3b6e6b406f15/src/app/core/state/state.service.ts#L64-L74

### move state stuff from rpc.service.ts into its own file (or in rpc-state)
These anything rpc-state related seriously doesn't belong there..
https://github.com/particl/particl-desktop/blob/a149e579cf512ba1d0bf50ff8cca3b6e6b406f15/src/app/core/rpc/rpc.service.ts#L131

rynomster:
We use shareReplay so that it can get the history. Get doesn't use observables so it doesn't need take.
The correct way would be to unsubscribe where it is no longer needed.

kewde:
re: get(), I see now :+1: 

I can't think of any situation in the codebase where we're actively using the history tho?

kewde:
@vikas-cis if you have done any changes on this issue, please push them.
I will do this.

pciavald:
@kewde do you want to integrate #741 here as well ? @HenkSwardt

kewde:
#797

PR #797, opened 2018-02-05
Rpc state refactor

kewde:


kewde:
#743

anandsinghparihar:
@kewde 

- [ ] Send all is not working (I guess some validation added in backed, so send all button should be display total_amount-transaction_fee).

rynomster:
By doing this, you're making everything that calls getwalletinfo update every time it's called, instead of if a state changes...

src/app/modals/encryptwallet/encryptwallet.component.ts
```diff
@@ -35,34 +36,32 @@ export class EncryptwalletComponent {
       this.log.d(`check password equality: ${password.password === this.password}`);
 
       if (this.password === password.password) {
-        this._rpc.state.set('ui:spinner', true);
+        this._rpcState.set('ui:spinner', true);
```
rynomster: This is ui state, not rpcState

kewde: I'm deleting every ui state from the rpcState
#769

src/app/core/state/state.service.ts
```diff
@@ -52,10 +52,16 @@ export class StateService {
     return state;
   }
 
-  observe(prop: string) {
+  observe(prop: string, subkey?: string) {
+    if (subkey) {
+      // TODO: maybe check if subkey exists?
+      // e.g observe('getblockchaininfo', 'blocks') will return only the 'blocks' key from the output.
+      return this._getObservablePair(prop).observable.map(key => key[subkey]);
```
kewde: If a subkey is provided in the observe(), then the observable.map(..) will only emit a value if the subkey changes.

```
observe('getblockchaininfo', 'blocks')
```
should only cause updates if blocks is updated.

Currently _every_ getwalletinfo/... will cause an update on 'blocks' whether it has actually changed or not..
`.distinctUntilChanged()` would fix this.

rynomster: It only emits if the value of blocks actually changes... Now it will always emit, on every single call

rynomster: Now you have to add a distinct on every observe, seems overkill.. It's still creating an observable for every key, so all this has accomplished is extra code and extra checks, using more cpu than before.

rynomster: You could return the observable with distinctUntilChanged(), as well as to possibly pass in the take until variable, which would reduce the code duplication everytime you use it.

kewde: @rynomster 
It was already emitting on every update, distinct or not. Take a look at the debug console, in the old code the 'blocks' were updating whether there was an actual change of value or not.. On top of it, it was replaying the whole stored set.

It uses the existing observables and returns a new one mapped to the existing one, I don't think cpu usage will change significantly.

adding distinctUntilChanged() :+1: 

We can't add `takeWhile(() => !this.destroyed)` here because that would only destroy the subscriptions if the _rpc state service is destroyed_, and not when the component is destroyed. The result would be the same as takeWhile(...) nowhere.

rynomster: You can pass in this.destroyed, that way it has context, and actually it was only replaying up until the last time it was used.. each time something is no longer used, it drops the history.
It was only emitting when the value changed. you can go put a debug message on master and you will see, only when a block changes, would it emit that the block has changed

PR #807, opened 2018-02-07
Switch from Electron to Muon

kewde:


package.json
```diff
@@ -39,17 +38,17 @@
     "compodoc": "compodoc -p tsconfig.json -d docs",
     "generate-release": "node ./modules/clientBinaries/generateVersions.js"
   },
-  "productName": "Particl Desktop",
+  "productName": "particl-desktop",
```
pciavald: why @kewde ?

pciavald:
screen shot 2018-02-12 at 1 42 32 pm

pciavald:
🎉  NO MORE BROKEN LAYOUT  🎉

kewde:
@pciavald 
Check if the particl-desktop folder is created, if not make it yourself.
If that's not the case, I'll have to add the fix for creating the log file, I think electron-log does this manually tho..

3n-mb:
@kewde I have a question, line 113 in main.js adds context isolation. Which contexts are these?
In electron, with preload, it is preload's context that gets separated from page's context. E.g. when preload sets window.foo = 4, it is present in a page in regular setting, and is missing when contexts' isolation is turned on.
But there is no preload in muon. What contexts are affected here?

(My fear here is that muon docs have leftovers from electron docs, making them a little misleading.)

kewde:
@3n-mb I replied to your issue re: contextIsolation on Muon, I suggest continuing there to make the information more accessible.

zaSmilingIdiot:
muon is no longer maintained, and no work has been done on the PR for years: pointless keeping it around still.

Issue #800, opened 2018-02-06
Replace angular2-qrcode


kewde:
It uses some native dependencies for some odd reason 'canvas', it's an optional dependency but it fails to build and has caused issue with electron-builder builds in the past, need to get rid of it.
Replace it with something more robust.

vikas-cis:
@kewde 
Can we try the below and use one of them if it make sense ? 

https://www.npmjs.com/package/ng2-qrcode
https://www.npmjs.com/package/angular2-qrcode-generator

Thanks

kewde:
https://www.npmjs.com/package/ng2-qrcode

That one looks fine to me.
Make sure to remove all the existing angular2-qrcode stuff, import the QrCodeModule in the CoreUiModule.

vikas-cis:
@kewde thanks will check and do a pr

vikas-cis:
@kewde I think we cannot use that 
https://github.com/PragmaticClub/ng2-qrcode/issues/12

Thanks

kewde:
Yes.

PR #808, opened 2018-02-08
replace angular2-qrcode with angularx-qrcode

vikas-cis:
#800


vikas-cis:
screen shot 2018-02-08 at 7 56 51 pm

PR #809, opened 2018-02-08
moved utl to core and change path accordingly

vikas-cis:


kewde:
Can you turn it into two tests: one for testnet addresses, one for mainnet addresses.
Also, please do some blackbox testing on the other classes in there.
Amount() is an easy one to test.

vikas-cis:
screen shot 2018-02-09 at 6 48 36 pm

This is how it looks like in mac, I think for current scenario we can just make it as it is and once we have electron 1.8.2 stable version comes up then will revert that and used the default behaviour.

vikas-cis:
@HenkSwardt writing test case for util classes 

Thanks

kewde:
@vikas-cis any update re: test cases?

vikas-cis:
@kewde yeah I am still working on it, I push the code once it done possibly tomorrow :)

PR #820, opened 2018-02-13
distinctUntilChanged()

kewde:


src/app/core/state/state.service.ts
```diff
@@ -53,12 +53,13 @@ export class StateService {
   }
 
   observe(prop: string, subkey?: string) {
+    let observable = this._getObservablePair(prop).observable.distinctUntilChanged();
     if (subkey) {
       // TODO: maybe check if subkey exists?
       // e.g observe('getblockchaininfo', 'blocks') will return only the 'blocks' key from the output.
-      return this._getObservablePair(prop).observable.map(key => key[subkey]);
+      return observable.map(key => key[subkey]);
```
rynomster: I think the property itself should be distinct, and not the entire node?

rynomster:
I think the property itself should be distinct, and not the entire node?

src/app/core/state/state.service.ts
kewde: My laptop battery was almost dead, I've moved the distinctUntilChanged to the last line so it works for subkeys/properties too.

The update check in the set() method (StateService) should provide distinctiveness for the node itself.
I find it weird why I was seeing doubles though.. The source code indicates that it should only emitted if it actually updated. I think the subscriptions not dying off might have been the cause for the 'duplicate' emits..

PR #825, opened 2018-02-14
fix mediantime issue

kewde:


vikas-cis:
@kewde seems no error on dev tool for median time.

PR #828, opened 2018-02-14
focus on input when debug console open

vikas-cis:
focus on input when debug console modal open


kewde:
Is this directive a copy paste from the modals? I think I've seen it before.

vikas-cis:
@kewde yeah I will do this common so it can be used where it is required

PR #839, opened 2018-02-15
Modal fix electron

rynomster:
Looks like its finally working.. css hacks :) The issue seems related to relative positioned elements inside a static positioned element.


vikas-cis:
@rynomster it works perfect to me on OSX .

Thanks

rynomster:
@allienworks when you get a chance could you have a look for us? :)

allienworks:
@rynomster seems to work fine for me. However I haven't this issue in this wallet/branch (I'm on Linux).

The main problem is, and I've said this so many times, that this is **not only** issue of modals, but everything that's animated. Check `market` branch, go to Buy and click through _Cart > Shipping > Confirm_.

rynomster:
Right now the most pressing issue is releasing the current release. Muon
doesn't work on Linux without a manual work around, and other things crash
on muon, so an electron release to get 1.1.0 out is the most obvious
release.. We can do the same fix for the other stuff, but chances are we
have some time to get it right before the MP release, and electron or Muon
also have some time.

allienworks:
Yeah, I just want us to keep in mind those other issues as well, regarding the future ahead. Agreed that the stepper etc. is not needed in 1.1, so we can definitely go with this fix for now.

Issue #836, opened 2018-02-15
Address book empty lable


bb2k16:
# Describe bug/issue

the empty labeled address under address book should follow the same format as the receive tab.

![image](https://user-images.githubusercontent.com/22429074/36255684-e8984816-1247-11e8-9132-7bf488be6e0f.png)
![image](https://user-images.githubusercontent.com/22429074/36255704-0186fbec-1248-11e8-82b1-51c837396d05.png)

# More information

- Operating system and version: Windows 10 x64
- Client version: RC5

vikas-cis:
@bb2k16 can you tell us, how you can get the label to be empty while adding the address?

bb2k16:
![image](https://user-images.githubusercontent.com/22429074/36257204-cb0ff356-124d-11e8-9ef6-53a8af5b0ff1.png)

Apologies, I am not able to reproduce the bug at this point, but as far as I could remember (according to the transaction history), those addresses were added automatically after I send funds to another address.

above screenshot was the VM and my local client sending back and forth, both produced have send 2 different transactions, and both got the address added to the address book automatically.

vikas-cis:
@bb2k16 thanks for the info, I can able to reproduce the issue :)

Thanks

PR #840, opened 2018-02-15
fix label issue and address book page update

vikas-cis:
#836


kewde:
Travis fail:
```
Error: src/app/wallet/wallet/shared/address-table/address-table.component.ts(69,26): error TS2341: Property '_addresses' is private and only accessible within class 'AddressService'.
```

PR #854, opened 2018-02-19
blocks to txcount for notification

vikas-cis:


vikas-cis:
@kewde I think it is required to have distinctUntilChanged thou why we need to remove that ?

Thanks

kewde:
distinctUntilChanged() is not required after #820

PR #862, opened 2018-02-22
market-api

kewde:
## Implement MarketService
The market service is very simple, nearly identical to the particl-core rpc service. 

## Implement MarketStateService
Same as the RpcStateService, might need some tweaking. 

## Categories
I've started work on the categories, we'll have to create a service to create/search/delete them. 

## Market & particl-core backend
Currently the market backend is pointing to a server of ludx, it can be used to experiment for now. 

> modules/webrequest/http-auth

Electron currently reroutes all HTTP calls and does a whitelist restriction on which domains can be accessed.
We intercept outgoing HTTP requests in electron and replace the Basic Authentication in the headers if the domain is in the whitelist.This means that the IPC isn't being used anymore by the RPC service.


src/app/core/rpc/rpc.service.ts
```diff
@@ -48,7 +49,7 @@ export class RpcService implements OnDestroy {
     private _http: HttpClient,
     private _ipc: IpcService
   ) {
-    this.isElectron = window.electron;
+    this.isElectron = false; // window.electron
```
kewde: Currently we're forcing electron over the old http request.

src/app/core/rpc/rpc.service.ts
```diff
@@ -86,10 +87,13 @@ export class RpcService implements OnDestroy {
...
-      headers.append('Content-Type', 'application/json');
-      headers.append('Authorization', 'Basic ' + btoa(`${this.username}:${this.password}`));
-      headers.append('Accept', 'application/json');
+
+      let headerJson = {
+       'Content-Type': 'application/json',
+       'Authorization': 'Basic ' + btoa(`${this.username}:${this.password}`),
+       'Accept': 'application/json',
+      };
+      let headers = new HttpHeaders(headerJson);
```
kewde: This needs to land in development too.
HttpHeaders can not be appended.

src/app/core/rpc/rpc.service.ts
rynomster: The reason we want it to go via electron is for security.. The transaction confirmation dialog will send supposed to pop up from the electron side, which currently isn't implemented

vikas-cis:
@HenkSwardt working with kewde on this branch :)

vikas-cis:
@kewde doesn't authenticate me to load the app ?

Thanks

src/app/core/rpc/rpc.service.ts
kewde: I know, we can do the same with this solution :) I just added the comment to keep track of this small hack.

kewde:
Fixed dev mode

PR #867, opened 2018-02-22
debounceTime on blocks

### LINKED ISSUE #794: GUI not synced
i've made a transaction, closed the wallet, opened it 2 hours later. I seen the daemon sync in the terminal, block height is at the latest block. However the transaction still shows 0 confirmations.
After 5min+ the confirmations count finally updated.
screenshot_20180203_034416

kewde:
Fix the throttle issue once and for all?
fixes #794


kewde:
fixes #794

PR #742, opened 2018-01-22
TX confirmation modal 2.0

AllienWorks:
- fixes #637 
- more previews for [blind TX](https://i.imgur.com/U1wEqBH.png) & [anon TX](https://i.imgur.com/Z1at8Hb.png)
- so far only static (= not working) layout → needs to be connected to the backend
- redesigned layout of TX confirmation modal
- added TX information (info about public/blind/anon TXs and their implications)
- added fee estimations
- added fiat countervalues (when ready to be implemented)

Note: this PR was based on @pciavald's https://github.com/particl/particl-desktop/pull/724 because I couldn't access the Send tab due to the bug on `dev`. If needed, I can rebase on `dev` when fixed.

### TODO

- [x] PRG: connect the static code to backend for live data (addresses, TX amounts etc.)

PR #903, opened 2018-03-18
fix parameters passed to core

kewde:


HenkSwardt:
Can we plesae confirm that this change has been added to #742 and has been merged?

kewde:
Not completely, still contains a change to the logging path for testnet.

Issue #881, opened 2018-03-09
MP Listings Page: Categories filter has no data


HenkSwardt:
The categories drop-down on the listings page needs to be linked to our categories list on the back-end as used on the “Add new listing” page.

See src/api/services/ItemCategoryService.findAll() for implementation of default category list. We need to keep in mind that new categories can be added and deleted, but users cant delete default list.

vikas-cis:
@HenkSwardt I gone through the latest stuff and seems categories are coming via the backend.

Thanks

vikas-cis:
I have the closed the above PR as it was showing some previous commits of my other PR due to which I am creating a new one 

Thanks

HenkSwardt:
I have tested and confirmed this to be working

PR #920, opened 2018-03-30
load all categories

vikas-cis:
Hye @HenkSwardt I have load all the categories on the listing page as before it was loading to one level due to which it doesn't filter the listing. There are duplicate categories have different id's so please ignore that.

Thanks


HenkSwardt:
https://github.com/particl/particl-desktop/issues/881

kewde:
This is insanity.

vikas-cis:
Hye @kewde thanks for your suggestion I have created another PR #921 , please let me know your view

Thanks

Issue #923, opened 2018-03-31
MP Listings - Base price returns 0 and breaks Cart


anandsinghparihar:
After creating bogus listing sometime I get an item without price, however, it won't be possible but this error occurs in my system. 
![issue 123](https://user-images.githubusercontent.com/17725557/38163129-34c4f5f4-350c-11e8-8b90-c1f0dfa6a53b.png)

PR #924, opened 2018-03-31
Base price handling in cart

anandsinghparihar:
#923

### DIFF
--- a/src/app/core/market/api/cart/cart.model.ts
+++ b/src/app/core/market/api/cart/cart.model.ts
@@ -12,13 +12,17 @@ export class Cart {
   get subTotal(): Amount {
     let total = 0.0;
     this.shoppingCartItems.map(shoppingCartItem => {
-      const object = shoppingCartItem.ListingItem;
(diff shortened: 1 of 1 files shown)


src/app/core/market/api/cart/cart.model.ts
```diff
@@ -30,8 +34,9 @@ export class Cart {
         shoppingCartItem.thumbnail = this.getThumbnail(object.ItemInformation.ItemImages[0]);
       }
       if (object.PaymentInformation) {
-        shoppingCartItem.integerPart = new Amount (object.PaymentInformation.ItemPrice.basePrice).getIntegerPart();
-        shoppingCartItem.fractionPart = new Amount (object.PaymentInformation.ItemPrice.basePrice).getFractionalPart();
+        const basePrice = this.getBasePrice(object.PaymentInformation);
+        shoppingCartItem.integerPart = new Amount (basePrice).getIntegerPart();
```
rynomster: Instantiate one `Amount` object instead of each time you need to use it

PR #938, opened 2018-04-04
load cart list when move again to the cart tab in buy page

### LINKED ISSUE #936: MP: adding Favs to Cart
When adding favourited item to Cart, it isn't added to Cart. There's a notification saying item was added, but it isn't there.

vikas-cis:
Fixes #936


AllienWorks:
Thanks @vikas-cis, this solves my issue. I noticed the items were in Cart, just the Cart didn't get updated. This works nicely.

Still needs code review.

PR #941, opened 2018-04-06
added parameter for default profile id in bid send

vikas-cis:
Added default profile id parameter in bid send

### DIFF
--- a/src/app/core/market/api/bid/bid.service.ts
+++ b/src/app/core/market/api/bid/bid.service.ts
@@ -26,7 +26,7 @@ export class BidService {
           this.market.call('bid', ['send', listing.hash, profile.id, profile.ShippingAddresses[0].id])
             .subscribe(
               (res) => {
-                this.log.d(`Bid placed for hash=${listing.hash} shipping to addressId=${addressIdOfProfile}`);
+                // this.log.d(`Bid placed for hash=${listing.hash} shipping to addressId=${addressIdOfProfile}`);
                 if (++nBidsPlaced === cart.listings.length) {
                   observer.next(true);
                   observer.complete();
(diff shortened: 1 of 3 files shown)


src/app/core/market/api/bid/bid.service.ts
```diff
@@ -20,12 +20,12 @@ export class BidService {
...
     let nBidsPlaced = 0;
     const addressIdOfProfile: number = profile.ShippingAddresses[0].id;
-
+    const profileId: string = profile.ShippingAddresses[0].profileId;
     return new Observable((observer) => {
       cart.listings.forEach((listing: Listing) => {
         if (listing.hash) {
           // bid for item
-          this.market.call('bid', ['send', listing.hash, addressIdOfProfile])
+          this.market.call('bid', ['send', listing.hash, addressIdOfProfile, profileId])
```
rynomster: the order is wrong..
It should be
`[send, listing_hash, profile_id, address_id]`

Also please just pass through `profile.ShippingAddresses[0].id` and `profile.Id` so that we're not duplicating variables unnecessarily at this point in time

vikas-cis:
@rynomster please have a look now 

Thanks

HenkSwardt:
@rynomster - Please review when you have a gap. 

Thanks

src/app/market/buy/checkout-process/checkout-process.component.ts
```diff
@@ -87,7 +87,7 @@ export class CheckoutProcessComponent implements OnInit {
 
   removeFromCart(shoppingCartId: number): void {
     this.cartService.removeItem(shoppingCartId).take(1)
-      .subscribe(this.getCart);
+      .subscribe(this.getCart.bind(this));
```
kewde: .bind(this) shouldn't be required afaik?

vikas-cis: it is required I guess as it doesn't bind the component if you called without that iirc

PR #947, opened 2018-04-18
PD-135 GUI - Hide ADD TO CART when the listing belongs to the seller

Skieter:


vikas-cis:
@Skieter it works nicely :) to me

kewde:
I suggest checking for the availability of the _template_. If the listing is linked to a template, then it is our own listing. This way it's checked, independently of the profile being used.

Issue #952, opened 2018-04-27
Market Gui - sidebar tabs wont work anymore after clicking "Buy"


SativaL:
# Describe bug/issue

Market gui - sidebar tabs wont work anymore

# How to reproduce
start the gui, click listings then click buy, then click sell.. it will crash

# More information

- Operating system and version: linux mint 18.x
- Client version: Client 1.1.0-RC1-alpha
Daemon 0.16.0.2

Gif: 

![out](https://user-images.githubusercontent.com/16978634/39383596-1d85fa3c-4a6a-11e8-9eea-ff1073681209.gif)

HenkSwardt:
Tested and confirmed that this has been resolved.

Thanks

PR #953, opened 2018-04-28
Hotfix for #952: sidebar tabs wont work anymore after clicking "Buy"

demyanrogozhin:
Just quickfix for JS exception, probably it would be better to find out why `this.selectedAddress` is not defined on view destruction.


demyanrogozhin:
Worth to mention:
this change fix visual issue, but error still present in communication with market rpc:
```
2018-04-29T20:42:59.698Z - debug: [:~/PART/particl-marketapi/controllers/RpcController]
controller.handleRPC(): 0=[profile address,update,,aaa,aaa,DEFAULT,aaa,,bbb,asdf,AL,123456]
2018-04-29T20:42:59.745Z - warn: [:~/PART/particl-marketapi/services/AddressService] Add
ress with the id=null was not found!
2018-04-29T20:42:59.756Z - info: [app] POST /api/rpc 404 55.047 ms - 125
```

PR #960, opened 2018-05-03
use yarn & builds

kewde:


kewde:
This pull request uses docker instead, also an upgrade version of NodeJS.
I haven't tested the builds, feel free to produce them.

Wine does work on this, but it is roughly 10 minutes slower than #976.

PR #965, opened 2018-05-03
Fix #963: MP Add groups in selectable categories filter drop-down (var. A)

### LINKED ISSUE #963: MP: add option groups in categories selector
This is the enhancement for selector drop-down with in Categories.

Expected:
As categories is hierarchy and inner categories should be visually different than upper-level categories.

Actual:
Currently categories displayed as flat list.

demyanrogozhin:
In categories filter drop-down list options aligned according level in
hierarchy (down to 3rd level, only two levels added now)


zaSmilingIdiot:
Closing due to age and that its currently being re-worked and re-implemented.

Issue #958, opened 2018-05-03
MP: Add/Edit item component - number fields allow negative numbers


demyanrogozhin:
Expected:
Number fields wouldn't allow negative numbers

Actual:
number fields allow negative numbers when i press Down key on keyboard or click down arrow

PR #969, opened 2018-05-06
MP: Add numeric input validation in "Add/Edit item" component #958

demyanrogozhin:


src/app/market/sell/add-item/add-item.component.ts
```diff
@@ -300,13 +302,13 @@ export class AddItemComponent implements OnInit, OnDestroy {
     this.log.d('Saving as a template.');
     if (this.templateId) {
       // update
-      this.update();
+      this.itemFormGroup.valid ? this.update() : this.snackbar.open('Invalid Listing');
```
kewde: I don't think this check is still required?
due to
```
+    if (this.itemFormGroup.invalid) {
+      return Observable.throw('Invalid Listing');
+    }
```

kewde:
LGTM otherwise :+1:

src/app/market/sell/add-item/add-item.component.ts
demyanrogozhin: This check is for updating existing template of not-yet-published listing, it calls `update` method *only* when form is valid.
Without this check user able to update existing template with invalid values.
Also at this point `update` itself need to be fixed - and this check could be moved into `update` method.
What do you think?

PR #1000, opened 2018-05-18
Market seller active listings issues

cis-shubham-t:


kewde:
Fix tests tho.

cis-shubham-t:
@kewde I ran the test cases, all they giving the positive result to me.

Issue #1010, opened 2018-05-21
Disable click on Shipping details if cart is empty.


HenkSwardt:
On the Checkout process, when the cart is empty, please remove the ability to click through to the second step (2 - Shipping Details)

PR #1013, opened 2018-05-22
Disable click on Shipping details if cart is empty. #1010

cis-shubham-t:


src/app/market/buy/checkout-process/checkout-process.component.ts
```diff
@@ -113,14 +129,16 @@ export class CheckoutProcessComponent implements OnInit, OnDestroy {
...
+    this.cartService.removeItem(shoppingCartId).take(1).subscribe(() => this.setCartItemCount());
   }
 
   clearCart(isSnack: boolean = true): void {
     this.cartService.clear().subscribe(res => {
       if (isSnack) {
         this.snackbarService.open('All Items Cleared From Cart');
       }
+      this.resetStepper();
+      this.setCartItemCount();
```
kewde: is this required, clearing the cart will trigger an update() and it will already call this.setCartItemCount() in the subscription.

src/app/market/buy/checkout-process/checkout-process.component.ts
```diff
@@ -167,6 +185,7 @@ export class CheckoutProcessComponent implements OnInit, OnDestroy {
   }
 
   clear() {
+    this.setCartItemCount();
```
kewde: Again, shouldn't be necessary?

src/app/market/buy/checkout-process/checkout-process.component.ts
```diff
@@ -113,14 +129,16 @@ export class CheckoutProcessComponent implements OnInit, OnDestroy {
   }
 
   removeFromCart(shoppingCartId: number): void {
-    this.cartService.removeItem(shoppingCartId).take(1).subscribe();
+    this.cartService.removeItem(shoppingCartId).take(1).subscribe(() => this.setCartItemCount());
```
kewde: I don't think it's needed here either.

src/app/market/buy/checkout-process/checkout-process.component.ts
```diff
@@ -79,7 +80,17 @@ export class CheckoutProcessComponent implements OnInit, OnDestroy {
 
     this.cartService.list()
       .takeWhile(() => !this.destroyed)
-      .subscribe(cart => this.cart = cart);
+      .subscribe((cart: Cart) => {
+        this.cart = cart
+
+        // no need to reset steper if cart is empty initially.
+        if (this.cart.countOfItems === 0 && !this.isInitialCartItemFetching) {
```
kewde: We can use `this.cart` to check if it's the initial fetching.

``` 
.subscribe((cart: Cart) => {
        // note: this.cart & cart, two different ones!
        if (this.cart && cart.countOfItems === 0) {
                 this.resetStepper();
        }
        this.cart = cart
        this.cartFormGroup.patchValue({ itemsInCart: this.cart.countOfItems });
} 
```

Still this seems a bit overkil and I don't trust the stepper.reset() I think we can probably simplify this.
Could you write out the issue again? What's exactly

kewde: Also this.stepper.reset () should set linear to true no?
Why is this.blockGoingBack needed

kewde: see #1020

cis-shubham-t: and `this.stepper.reset ()` set linear false and default.

HenkSwardt:
https://github.com/particl/particl-desktop/issues/1010

PR #1014, opened 2018-05-22
error handling while bidding on order according to scenarios

vikas-cis:
GUI error handling for "No unspent outputs" and "Insufficient funds"

If this is returned:

2018-05-22T09:17:08.315Z - warn: [api\services\BidActionService] You are too broke...
2018-05-22T09:17:08.318Z - info: [app] POST /api/rpc 404 129.336 ms - 81

Message should display as follow:

Insufficient funds to place the order.

And for this error:

2018-05-22T09:28:50.170Z - warn: [api\services\BidActionService] No unspent outputs
2018-05-22T09:28:50.173Z - info: [app] POST /api/rpc 404 87.016 ms - 77

Please display:

No unspent outputs available. Please complete active locked escrow.

### DIFF
--- a/src/app/core/market/api/bid/bid.service.ts
+++ b/src/app/core/market/api/bid/bid.service.ts
@@ -8,6 +8,10 @@ import { Cart } from 'app/core/market/api/cart/cart.model';
 import { Listing } from 'app/core/market/api/listing/listing.model';
 import { Bid } from 'app/core/market/api/bid/bid.model';
 
+export enum errorType {
+  unspent = 'No unspent outputs available. Please complete active locked escrow.',
+  broke = 'Insufficient funds to place the order.'
+}
 
 @Injectable()
 export class BidService {
@@ -31,7 +35,10 @@ export class BidService {
                   observer.next(true);
                   observer.complete();
                 }
-              }, observer.error);
+              }, (error) => {
+                error = error.includes('unspent') ? errorType.unspent : errorType.broke
+                observer.error(error)
+              });
         }
       });
     });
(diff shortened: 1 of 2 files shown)


HenkSwardt:
Testing Errors. 

Scenario 1:

Try to purchase an item costing more than your available funds will produce this on screen

image
 
Scenario 2: 

Placed a bid on an item and trying to bid on a second item with a new wallet. The following message will be produced

image

src/app/core/market/api/bid/bid.service.ts
```diff
@@ -31,7 +35,10 @@ export class BidService {
                   observer.next(true);
                   observer.complete();
                 }
-              }, observer.error);
+              }, (error) => {
+                error = error.includes('unspent') ? errorType.unspent : errorType.broke
```
kewde: Fix the logic here please.
It will always output 'broke' if it's not the unspent error, there might be more errors that we're not aware of!

```
if(error.includes('unspent')) {
  error unspent
} else if(error.includes('the string to match the broke error')) {
  error broke
} else {
  just send the error message!
}

```

vikas-cis: @kewde so you believe there will be more errors accept the above ?

Thanks

vikas-cis: @kewde what will be the message if other than this we need to show ?

thanks

src/app/core/market/api/bid/bid.service.ts
```diff
@@ -31,7 +36,16 @@ export class BidService {
...
                   observer.complete();
                 }
-              }, observer.error);
+              }, (error) => {
+                if (error.includes('unspent')) {
+                  error = errorType.unspent;
+                } else if (error.includes('broke')) {
+                  error = errorType.broke;
+                } else {
+                  error = errorType.other;
```
kewde: Return the original error.

kewde:
Test didn't start but it's ok.

PR #1016, opened 2018-05-23
electron 2.0.1 & progress for updater

kewde:


main.js
```diff
@@ -96,7 +96,7 @@ function initMainWindow() {
       webviewTag: false,
       nodeIntegration: false,
       sandbox: true,
-      contextIsolation: true,
+      contextIsolation: false,
```
kewde: For some reason contextIsolation: true will ruin attaching things to the window from the preloader.

PR #1047, opened 2018-06-06
fixed double quote and single quote values - https://particl.atlassian.net/browse/PD-154

vikas-cis:
@kewde 
1)'addnode IPADDRESSHERE add'
2)walletsettings changeaddress "{}"

When wallet is locked we got http internal server error but if you unlocked the wallet then it will work.
Updated : Due to this reason we have added 'Method not found' message

Thanks


kewde:
Move the parser into a different function and write a test suite.

PR #1052, opened 2018-06-07
Update develop (no market)

kewde:


cryptoguard:
@kewde Can we get a new build? Link has expired.

PR #1055, opened 2018-06-08
close preview listing modal after add to cart -https://particl.atlassian.net/browse/PD-138

vikas-cis:


src/app/market/listings/preview-listing/preview-listing.component.html
```diff
@@ -78,6 +78,6 @@
   </div>
   <div class="right" fxLayout fxLayoutAlign="end center">
     <app-favorite [listing]="data?.listing" [detail]="true"></app-favorite>
-    <app-add-to-cart [listing]="data?.listing" [detail]="true"></app-add-to-cart>
+    <app-add-to-cart [listing]="data?.listing" [detail]="true" (onClose)="dialogClose()"></app-add-to-cart>
```
kewde: onAdded

src/app/core/market/market.service.ts
```diff
@@ -41,7 +41,7 @@ export class MarketService {
     return this._http.post(this.url, postData, { headers: headers })
         .map((response: any) => response.result)
         .catch((error: any) => {
-
+          // Inconsistency when you do add to cart
```
kewde: More explanation?

vikas-cis: we are getting 404 error while try to add same item in the cart and got error something like `error.error.error`, Due to which we got `[object object]` in the error message.

kewde: Then fix the error handling..

kewde: The in-code comment here is useless.

vikas-cis: ok

PR #1079, opened 2018-06-22
+build

vikas-cis:


kewde:
You can only trigger builds from branches within the particl-desktop repository, not from a different remote.
```
vikas-cis wants to merge 1 commit into particl:market from vikas-cis:vikas/build
```

Should be: 
```
vikas-cis wants to merge 1 commit into particl:market from particl:vikas/build
```

PR #1088, opened 2018-07-05
Created select-search component to fixed Select box filter issue.

cis-shubham-t:
- Currently Added the select-search component on listing page.


cis-shubham-t:
@AllienWorks  
Can you please have a look on the design related breaks listing page.

AllienWorks:
Few things:
- I've made the selects a bit wider, so there's not that much cropped text in its options
- I've removed the placeholders from the selects, it's better this way
- @cis-shubham-t: the Categories select displays initially only "All categories" - you have to go to other page and back to see all of them - can we load them dynamically?

kewde:
Please review first @AllienWorks

src/app/core-ui/material/mat-otpgroup-select-search/mat-otpgroup-select-search.component.ts
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  private _filterGroup(value: string): any[] {
+    if (!this.options || !this.options.length) {
+      return []
+    }
+
+    if (value) {
+      return this.options
+        .map(group => ({ name: group.name, subCategoryList: this._filter(group.subCategoryList, value) }))
```
kewde: This has to be a completely generic piece of code, it shouldn't have any references to subCategoryList please change the name.

src/app/core/market/api/category/category.service.ts
```diff
@@ -38,4 +38,18 @@ export class CategoryService {
   search(searchString: string): Observable<any> {
     return this.market.call('category', ['search', searchString]);
   }
+
+  getCategoryByCategoryId(categories: any[], categoryId: number)  {
```
kewde: Hmmm, this seems a bit clunky..
Why is this required? And can we get the categories through list(à-) instead of passing them?

kewde: I can't even ctrl+f find the usage of this function, is this dead code?

AllienWorks:
I will make some layout changes.. In the meantime:

> @cis-shubham-t I've added the search-component on the "Add-edit listing" and "shipping-address-form" (checkout-process) component there are some design changes required can you please have a look on the same.

**Add/edit listing**
There seem to be some issue with Category select - it doesn't show all the available options when you click on it. As you can see, the Country select is working fine, but Category does nothing:

bug-select_search

**Shipping address form**
This works OK.

AllienWorks:
Regarding the core stuff from this PR (the `select-search` component), that is done and working nicely.
I definitely approve that.

However these is still an issue with Category select:
- on **Add/edit listing** page, it shows just `Loading...` indefinitely. User has to type something and then delete it all to reveal all the available options. We need to load them right away (not just after user's input)
- on **Listings** page, after clicking the select, only `All categories` shows up (as it was typed and filtered by user). To reveal all available categories, user has to delete the `All categories` text first.. This should be fixed so it works the same as with e.g. Countries (= display all available countries when user clicks the input)

cis-shubham-t:
@AllienWorks @kewde 
I did some modification, I guess that with make the category search-select box and both category and country working same for me.

@AllienWorks can you please have a look on your end.

thanks :)

AllienWorks:
There should be also some kind of check for validating the input.

Example: user wants to select `Furniture` category, so starts typing "Fur.." and the `Furniture` option pops up. But instead of selecting it, user might just continue writing and accidentaly makes a typo, so ends up submitting `Furnittur` (or whatever).

We need to make sure users don't just write their own "custom" categories and that their input matches one of the options available. I guess the easiest way would be to implement the [error messages on invalid inputs](https://material.angular.io/components/select/overview#changing-when-error-messages-are-shown)?

kewde:
We'll fix the issue in another bug report, going to merge as the component is pretty good overall.

PR #1089, opened 2018-07-06
Voting

vikas-cis:
Voting and Devoting stuff


src/app/core/market/api/report/report.service.ts
```diff
@@ -0,0 +1,36 @@
...
+    private profile: ProfileService
+  ) {
+
+    this.profile.default().takeWhile(() => !this.destroyed).subscribe((prof: any) => {
+      this.defaultProfileId = prof.id;
+    });
+  }
+
+  post(listing: Listing) {
+    console.log(listing.hash);
```
zaSmilingIdiot: Can we either change this to a debug log message, or remove this statement please...

src/app/market/shared/report/report.component.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    dialogRef.componentInstance.isConfirmed.subscribe((res: any) => {
+      this.modals.unlock({timeout: 30}, (status) => this.reportItem());
+    });
+  }
+
+  reportItem(): void {
+    this.reportService.post(this.listing).subscribe(report => {
+      this.listing.isFlagged = !this.listing.isFlagged;
+      this.snackbar.open(`Listing ${this.listing.title} Flagged Successfully`);
```
zaSmilingIdiot: Can we change the text here to be instead something like `Listing ${this.listing.title} has been reported successfully`

src/app/market/shared/report/report.component.html
```diff
@@ -0,0 +1,14 @@
+<mat-icon *ngIf="flag === false" fontSet="partIcon" fontIcon="part-flag"
+        [ngClass]="{ 'active': listing?.isFlagged }"
+        [color]="listing?.isFlagged ? 'warn' : 'basic'" 
+        [matTooltip]="listing?.isFlagged ?  'Remove from Report' : 'Report'"
+        (click)="toggle()"></mat-icon>
```
zaSmilingIdiot, line 7: The 'Remove from Report' tooltip, along with the clickable action of the icon when a item has been flagged suggests to the user that some user action is available when it is not.

I suggest possibly changing the 'Remove from Report' tooltip  to something like "this item has been reported for inappropriate content") when a listing item has been flagged. And then possibly displaying a flat, non-clickable icon or button. I realize that the toggle() function does include a check for this (and that could probably stay in place as a redundant check), but the icon/button itself should not be clickable once the item is flagged.

src/app/market/shared/report/report.component.html
```diff
@@ -0,0 +1,14 @@
...
+
+<button *ngIf="flag === true" mat-button class="small"
```
zaSmilingIdiot: Same comment as above

src/app/market/shared/report/report.component.html
vikas-cis-zz: yeah we are looking for the correct wording as we don't use unflagging stuff

src/app/core/market/api/report/report.service.ts
vikas-cis-zz: yup

src/app/market/shared/report/report.component.html
vikas-cis-zz: probably a class that comes after it gets flagged @AllienWorks can you please have a look and suggest a better that we can replace after flagging so user don't get idea of clicking on it

src/app/core/market/api/report/report.service.ts
vikas-cis-zz: and I think that console is even not needed their :) so I am going remove it 👍

src/app/market/shared/report/report.component.html
AllienWorks: @vikas-cis Let's add a new class when icon was clicked (something like `active` or `reported`), I will style it afterwards.

If we could make it disabled, that would be perfect – you can wrap the icon in small button for this:
```
<button mat-button class="small">
  <mat-icon fonticon="part-xxxx" fontset="partIcon"></mat-icon>
</button>
```

vikas-cis:
@AllienWorks there is already active class available so I guess you can use that to style it more.

Thanks

zaSmilingIdiot:
From external discussion, this still requires a means (in the UI) for a user to submit a vote on a flagged item. That is, to be able to vote to keep or remove an item that has been flagged.

HenkSwardt:
Hey @zaSmilingIdiot 

Just to clarify my understanding. :)

We want to add the ability for a user to vote directly on the listing instead of going to the proposals page and voting?

So, in short. If user a flagged an item as inappropriate, he will automatically cast the first vote to remove. (At this stage the listing only has a flag item icon/button). Once submitted, users b, c, d .... z, will now see that the item has been flagged and should have buttons visible to cast their vote directly from the listing page. ex "Keep Listing" & "Remove Listing". (These will only be visible on reported listings.

Please confirm my understanding is correct. 

Thank you

zaSmilingIdiot:
> Hey @zaSmilingIdiot
> 
> Just to clarify my understanding. :)
> 
> We want to add the ability for a user to vote directly on the listing instead of going to the proposals page and voting?
> 


Thats exactly right. Your example also covers one of the proposed user flow for reporting listing items. Currently, the reported/flagged items are not included in the 'Proposals' section. So there is no current place for users to cast votes on a flagged/reported listing.

One option is to include the reported listing item proposals in the current 'Porposals', but as discussed previously, this leads to chaos in the proposals section - there's the potential for one of the proposal types (community proposals and reported listing items) to drown the other out.

Another possibility is to create a separate 'Reported Items' section (the exact name to be discussed), which contains the proposals for just the reported listing items.

The third option is as per your example, in that the option to vote is included on the listing item itself (not in the listings page, but in the item detail section).

There may be additional options, but the above 3 seem to cover the broader options that we have available. So as for the 3 options, I tend to agree with @xludx 's suggestion that a user should probably want to see the entire listing in order to decide whether to vote on it or not. For example, perhaps one of the pictures, or some wording in the long description resulted in it being flagged. This becomes more intuitive to vote on if the voting options for such a flagged item are included in the listing section itself, which leans towards the 3rd suggested option above.

AllienWorks:
I agree that the voting should be on Listing's detail, that makes most sense. The flagged icon should be visible on Listing's thumbnail (not in its detail) as well, to indicate that there is something "wrong".

Regarding the voting process, we could make use of [Bottom sheet](https://material.angular.io/components/bottom-sheet/overview) (with "Report item" vs "Keep item"; we could include more of the options like "This is illegal in my country" etc.) after clicking the Report button.

And to avoid triggering more than 1 voting proposal per Listing (as discussed, when multiple people will simultaneously click on Report button), couldn't we automatically create these voting proposals each time new Listing gets submitted by seller? Just an idea, even though it might unnecessarily spam the network..?

zaSmilingIdiot:
> Regarding the voting process, we could make use of [Bottom sheet](https://material.angular.io/components/bottom-sheet/overview) (with "Report item" vs "Keep item"; we could include more of the options like "This is illegal in my country" etc.) after clicking the Report button.
> 

I think we only get 2 options to vote on at the moment, as per the marketplace, and I'm in favour of leaving it that way: the voting process is a binary decision (keep the listing item or not). In saying that, I think having feedback on why a user is voting a particular way might be helpful to the listing item creator. That might need to be a separate additional "plugin" to the process though (even if later the UI makes it appear as one action/activity). 

> And to avoid triggering more than 1 voting proposal per Listing (as discussed, when multiple people will simultaneously click on Report button), couldn't we automatically create these voting proposals each time new Listing gets submitted by seller? Just an idea, even though it might unnecessarily spam the network..?

I think this has already been solved in the marketplace, in which they drop the later requests to flag an item if the item has already been flagged previously. 

The one thing that we can validate with the marketplace code is whether those additional requests that are being dropped count towards the vote to remove the item or not (additional proposals should not created, but technically the additional flag requests should count towards the vote to remove the item). That's something related to the marketplace though, so will have to raise that query there.

HenkSwardt:
Thanks all for the input.

@zaSmilingIdiot 
@AllienWorks 
@xludx 
@vikas-cis

AllienWorks:
- display reported items on Listings overview (Reported icon in top right)
- position of Report button in detail



- when Listing gets reported, display this voting "stripe" & inform user about what's going on
- voted option becomes "activated" (e.g. colored), the other gets disabled

vikas-cis-zz:
@AllienWorks I feel like I have done all the changes related to coding and now you can feel free to do css changes :)

Thanks

PR #1090, opened 2018-07-06
Accept terms

vikas-cis:
Accept terms and Condition modal


vikas-cis:
@AllienWorks can you please check the modal and if need you can design 

Thanks

src/app/core/terms/terms-txt.ts
```diff
@@ -0,0 +1,3 @@
+export const termsObj = {
+  version: '0.1'
```
kewde: Can we store the text in here?

vikas-cis: @kewde Is there any reason for that ? I believe we can make the text static in html and as we have versioning stuff so whenever we change in text we can just update the version ?

Thanks

kewde: The text and its version number form one entity, so it would be nice to have them next to eachother so we don't forget to update the version number.

kewde:
Sorry my bad, I didn't see that this was just a modal, please move it to the modals directory, thanks.

AllienWorks:
Travis agrees that one long line of text is a no-go :)
```
ERROR: /home/travis/build/particl/particl-desktop/src/app/modals/terms/terms-txt.ts[3, 1]: Exceeds maximum line length of 140
```

AllienWorks:
The `Decline` button doesn't work for me (does nothing). Nothing even in Console.. (@Linux)

PR #1094, opened 2018-07-11
Orders: sorting/filtering

AllienWorks:
screenshot from 2018-07-11 13-35-06

- enable sorting/filtering of Orders
- more at https://particl.atlassian.net/browse/PD-229

@vikas-cis - ready for implementation :)


src/app/market/shared/orders/orders.component.ts
```diff
@@ -93,6 +102,16 @@ export class OrdersComponent implements OnInit, OnDestroy {
     )
   }
 
+  filter(): void {
+    console.log(this.filters.status);
+    this.loadOrders();
+  }
+
+  clear(): void {
+    this.filters.status = '*';
```
kewde: This seems pretty fragile, when removing the filters, clear() must always be called.

When doing the call, can't you just
```
this.bid.search(this.profile.address, this.type, this.filters.status || '*' , this.filters.search || '')
```

Default values should be passed like that.

kewde: Perhaps turn it into a get():

```
get filterStatus() {
  return this.filters.status || '*'
}

Comment (kewde) [reply src/app/market/shared/orders/orders.component.ts]: idem for search

File: src/app/market/shared/orders/orders.component.ts [hunk 2]
```diff
@@ -72,12 +76,17 @@ export class OrdersComponent implements OnInit, OnDestroy {
...
       .take(1)
       .subscribe(bids => {
         console.log('called >>>>>>>>>>>>>>>>>', bids);
         // Only update if needed
         if (this.hasUpdatedOrders(bids.filterOrders)) {
+          // Initialize model only when its fetching for all orders.
+          if (this.filters.status === '*') {
+            this.order_filters = new OrderFilter();
+          }
+          this.order_filters.setOrderStatusCount(this.filters.status, bids.filterOrders)
```
kewde: here too

src/app/core/market/api/bid/bid.service.ts
```diff
@@ -43,8 +43,9 @@ export class BidService {
     return 'Placed all orders!';
   }
 
-  search(address: string, type?: any): Observable<any> {
-    const params = ['search', '*', '*', 'ASC'];
+  search(address: string, type?: any, status?: string, search?: string): Observable<any> {
```
kewde: return type missing

src/app/market/shared/orders/orders.component.ts
vikas-cis: I think we have default is already their so we can use that as well ?

PR #1102, opened 2018-07-24
Display cart price according to select country from the location selector (price OR price + shipping price)

cis-shubham-t:
- Display the cart total according to the selected country from the location selector.

- display the section shipping price option at the checkout process item listing for the first process, if the country is selected from the location.

### DIFF
--- a/src/app/core-ui/material/mat-select-search/mat-select-search.component.ts
+++ b/src/app/core-ui/material/mat-select-search/mat-select-search.component.ts
@@ -1,4 +1,12 @@
-import { Component, Input, Output, EventEmitter, OnChanges, OnInit } from '@angular/core';
+import {
+  Component,
+  Input,
+  Output,
+  EventEmitter,
+  OnChanges,
+  ViewChild,
+  ElementRef,
+  OnInit } from '@angular/core';
 import { FormControl } from '@angular/forms';
 import { Observable } from 'rxjs';
 import { map, startWith } from 'rxjs/operators';
@@ -11,6 +19,7 @@ import { CountryListService } from 'app/core/market/api/countrylist/countrylist.
   styleUrls: ['./mat-select-search.component.scss']
 })
 export class MatSelectSearchComponent implements OnInit, OnChanges {
+  @ViewChild('textInput') textInput: ElementRef;
   @Input() public showValueOf: string = 'name'; // default key as 'name'.
   @Input() public placeHolder: string = '';
   @Input() public options: any[] = [];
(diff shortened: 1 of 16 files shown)


cis-shubham-t:
@AllienWorks 
The dot (.) issue fixed now can please have a look :)  

fixed

cis-shubham-t:
@AllienWorks 
@HenkSwardt 

I've added the mechanism to remove the focus from the input box when the user selects an option from the available options list.
That's resolved our issue related with the location selector and other search select box.

Please have a look :)

AllienWorks:
> There's a small issue of Shipping price in Cart - it's not displaying the . in price.

Yes, this has been resolved. Good job!

> I've added the mechanism to remove the focus from the input box when the user selects an option from the available options list.

Tested & works as needed.

**Still needs code review.**

src/app/market/listings/listings.component.ts
```diff
@@ -79,7 +79,11 @@ export class ListingsComponent implements OnInit, OnDestroy {
     private countryList: CountryListService
   ) {
     console.warn('overview created');
-    this.selectedCountry = this.listingService.cache.selectedCountry;
+    if (this.listingService.cache.selectedCountry) {
```
kewde: Why just not make it a type country in the post-listings cache?

src/app/core-ui/material/mat-select-search/mat-select-search.component.ts
```diff
@@ -59,15 +68,20 @@ export class MatSelectSearchComponent implements OnInit, OnChanges {
     return this.options.filter(option => option[this.showValueOf].toLowerCase().includes(filterValue));
   }
 
-  public _onBlur($event: any): void {
-    const val = $event.target.value;
+  public _onBlur(): void {
```
kewde: Why was $event not working?

cis-shubham-t: yes, that's working fine.
But in the same component, we have the ``Viewchild`` (native element) ref of the <input-field>.
that's the reason I'm replaced ;

``` $event.target = this.textInput.nativeElement  ```

**Comment by AllienWorks** [comment]
LGTM

File: src/app/market/listings/preview-listing/preview-listing.component.html [hunk 3]
```diff
@@ -42,9 +42,9 @@
         </td>
       </tr>
 
-      <tr *ngIf="listingServiceCache?.selectedCountry">
+      <tr *ngIf="listingServiceCache?.selectedCountry?.iso">
```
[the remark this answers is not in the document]
cis-shubham-t: Yes, working in that case as well.

Its just to added for make condition more specfic.

PR #1108, opened 2018-07-31
Fixed multiple image upload issue and disabled "publish and save"  button while the listing adds process.

cis-shubham-t:
- Fixed multiple image upload issue and disabled "publish and save"  button while the listing adds process.
- Added a hack for maintaining the multiple image upload functionality.

### DIFF
--- a/src/app/core/market/api/template/image/image.service.ts
+++ b/src/app/core/market/api/template/image/image.service.ts
@@ -29,14 +29,25 @@ export class ImageService {
   public upload(template: Template, images: Array<any>): Promise<Template> {
     let nPicturesAdded = 0;
 
+    /* @TODO
+     * remove that shity hack once multiple image upload functionality got refactured from the backend.
+     * related `totalnPicturesAdded` stuff will be remove in that case.
+     */
+    let totalnPicturesAdded = images.length;
     return new Promise((resolve, reject) => {
       images.map(picture => {
         this.log.d('Uploading pictures to templateId=', template.id);
         this.add(template.id, picture).take(1).subscribe(res => {
-          if (++nPicturesAdded === images.length) {
+          this.log.d(`image uploaded`, nPicturesAdded)
+          if (++nPicturesAdded === totalnPicturesAdded) {
             this.log.d('All images uploaded!');
             resolve(template);
           }
+        }, error => {
+          // at least we have some images uploaded and we are assuming it as resolving it as a success
+          --totalnPicturesAdded;
+
+          this.log.d(`error in image upload ${template.id}`, nPicturesAdded)
         });
 
       });
--- a/src/app/market/sell/add-item/add-item.component.ts
+++ b/src/app/market/sell/add-item/add-item.component.ts
@@ -356,12 +356,13 @@ export class AddItemComponent implements OnInit, OnDestroy {
     this.upsert().then(t => {
       this.modals.unlock({timeout: 30}, (status) => {
         this.template.post(t, 1)
-        .finally(() => this.isInProcess = false)
-        .toPromise().then(listing => {
+        .subscribe(listing => {
           this.snackbar.open('Succesfully added Listing!')
           this.log.d(listing);
           this.backToSell();
-        });
+        },
+        (error) => this.log.d(error),
+        () => this.isInProcess = false)
       });
     }, err => this.snackbar.open(err));
   }
--- a/src/app/market/sell/add-item/add-item.component.ts
+++ b/src/app/market/sell/add-item/add-item.component.ts
@@ -47,6 +47,7 @@ export class AddItemComponent implements OnInit, OnDestroy {
 
   selectedCountry: Country;
   selectedCategory: Category;
+  isInProcess: boolean = false;
 
   constructor(
     private router: Router,
@@ -337,6 +338,7 @@ export class AddItemComponent implements OnInit, OnDestroy {
     if (!this.validate()) {
       return;
     };
+    this.isInProcess = true;
     this.log.d('Saving and publishing the listing.');
     this.publish();
   }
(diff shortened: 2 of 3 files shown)


src/app/market/sell/add-item/add-item.component.ts
```diff
@@ -354,12 +356,16 @@ export class AddItemComponent implements OnInit, OnDestroy {
     this.upsert().then(t => {
       this.modals.unlock({timeout: 30}, (status) => {
         this.template.post(t, 1).toPromise().then(listing => {
-        this.snackbar.open('Succesfully added Listing!')
-        console.log(listing);
-        this.backToSell();
+          this.isInProcess = false;
+          this.snackbar.open('Succesfully added Listing!')
+          console.log(listing);
```
zaSmilingIdiot: Can this be wrapped instead with a debug log line if it is necessary to output the listing information?

src/app/market/sell/add-item/add-item.component.ts
```diff
@@ -354,12 +356,16 @@ export class AddItemComponent implements OnInit, OnDestroy {
...
+          this.backToSell();
         });
       });
-    }, err => this.snackbar.open(err));
+    }, err => {
+      this.isInProcess = false;
+      this.snackbar.open(err)
```
zaSmilingIdiot: Do we not want to rather output a more user friendly error message here if this fails?

src/app/core/market/api/template/image/image.service.ts
```diff
@@ -29,14 +29,25 @@ export class ImageService {
...
 
+    /* @TODO
+     * remove that shity hack once multiple image upload functionality got refactured from the backend.
+     * related `totalnPicturesAdded` stuff will be remove in that case.
+     */
+    let totalnPicturesAdded = images.length;
     return new Promise((resolve, reject) => {
       images.map(picture => {
         this.log.d('Uploading pictures to templateId=', template.id);
         this.add(template.id, picture).take(1).subscribe(res => {
```
kewde: Can't we Promise.all() this? and turn this toPromise()

kewde: You can catch the errors with .catch() then .finally() to resolve/reject?

src/app/market/sell/add-item/add-item.component.ts
cis-shubham-t:
```
https://github.com/particl/particl-desktop/blob/market/src/app/core/market/market.service.ts#L46
```
We have already handled the error in the service regarding the message and if backend return differently then at least we have that error to return in the same way as what it is.

zaSmilingIdiot: I agree, having the error is helpful. But if we could potentially be printing errors that include line numbers, indications of things being undefined, etc, then those types of error messages may not be helpful to a user (non-developer user type of person). If the error message is vague, it does not indicate what actually did go wrong, for example: 
was the listing successfully published but there was a problem changing the view, OR
did the publish fail because they had some invalid value, OR
did the publish fail because of a connectivity issue, 
etc

Perhaps prefixing the error message with something indicating that the listing failed to be published would be a little more helpful to the user than just presenting an error message.In other words, an error message needs to tell the user what went wrong but at the same time hint at what action they should try to perform next. In this case, a 'Failed to publish your listing' type of error suggests that the subsequent action may be to try republish the listing again (it is explicitly known that the previous attempt failed, rather than the user needing to first verify whether it possibly succeeded or not).

cis-shubham-t: @zaSmilingIdiot  Now your call on the same?

src/app/market/sell/add-item/add-item.component.ts
```diff
@@ -354,12 +356,16 @@ export class AddItemComponent implements OnInit, OnDestroy {
     this.upsert().then(t => {
       this.modals.unlock({timeout: 30}, (status) => {
         this.template.post(t, 1).toPromise().then(listing => {
-        this.snackbar.open('Succesfully added Listing!')
-        console.log(listing);
-        this.backToSell();
+          this.isInProcess = false;
```
kewde: on complete:
```
this.isInProcess = false;
```

src/app/market/sell/add-item/add-item.component.ts
```diff
@@ -354,12 +356,16 @@ export class AddItemComponent implements OnInit, OnDestroy {
...
+          this.snackbar.open('Succesfully added Listing!')
+          console.log(listing);
+          this.backToSell();
         });
       });
-    }, err => this.snackbar.open(err));
+    }, err => {
+      this.isInProcess = false;
```
kewde: rm

cis-shubham-t:
Hey @HenkSwardt 
As per discussion with @kewde we have to change and refactor a bit related to multiple image upload so on the basis of that can we have a new task for it?

Thanks

kewde:
@cis-shubham-t please make it use on complete.

HenkSwardt:
Logged a task to address the new feature in a new PR:

https://particl.atlassian.net/browse/PD-308

cis-shubham-t:
@kewde 
I added the .finally() as we discussed.

src/app/market/sell/add-item/add-item.component.ts
```diff
@@ -353,10 +355,12 @@ export class AddItemComponent implements OnInit, OnDestroy {
   private async publish() {
     this.upsert().then(t => {
       this.modals.unlock({timeout: 30}, (status) => {
-        this.template.post(t, 1).toPromise().then(listing => {
-        this.snackbar.open('Succesfully added Listing!')
-        console.log(listing);
-        this.backToSell();
+        this.template.post(t, 1)
+        .finally(() => this.isInProcess = false)
```
kewde: After toPromise()?

PR #1113, opened 2018-08-10
Fixes minor multiwallet issues

zaSmilingIdiot:
Specifically fixes the following issues:

* the wallets that were not displayed on the wallet sidebar menu are now displayed;
* Reduced the number of steps in the Create/Restore wallet workflow, to 5, as the process exists after the 5th step (thus never getting to the encrypt wallet step) as the remainder steps are not currently implemented;
* The create wallet process now displays an inline error message if the name of the wallet being created already exists.
* Changed the 'Encrypt wallet' button on the first step of the create/restore wallet process to rather be a span, as the button has no associated functionality (and doesn't appear that it should have at this particular step).
* In the restore wallet > Enter Mnemonic step, an invalid mnemonic was previously displaying a blank white screen, but now correctly returns to the Mnemonic input screen (with an error)

The above are using the 0.17.0.2-alpha version of particl-core daemon.

I cannot seem to reproduce the issue: 
* New wallet creation: `wallet gets stuck (doesn't load TXs, balances, ..), 2 cold staking widgets are shown, other pages not working/loading as well`.

I also seem to be using different code formatting settings, so if the noise of the formatting changes are a problem, I can fix this as well.

### DIFF
--- a/main.js
+++ b/main.js
@@ -1,11 +1,10 @@
-const electron      = require('electron');
-const app           = electron.app;
+const electron = require('electron');
+const app = electron.app;
 const BrowserWindow = electron.BrowserWindow;
-const path          = require('path');
-const fs            = require('fs');
-const url           = require('url');
-const platform      = require('os').platform();
-
+const path = require('path');
+const fs = require('fs');
+const url = require('url');
+const platform = require('os').platform();
 
 /* correct appName and userData to respect Linux standards */
 if (process.platform === 'linux') {
@@ -14,16 +13,15 @@ if (process.platform === 'linux') {
 }
 
 /* check for paths existence and create */
-[ app.getPath('userData'),
-  app.getPath('userData') + '/testnet'
-].map(path => !fs.existsSync(path) && fs.mkdir(path));
+[app.getPath('userData'), app.getPath('userData') + '/testnet'].map(
+  path => !fs.existsSync(path) && fs.mkdir(path)
+);
 
-if (app.getVersion().includes('RC'))
-  process.argv.push(...['-testnet']);
+if (app.getVersion().includes('RC')) process.argv.push(...['-testnet']);
 
 const options = require('./modules/options').parse();
-const log     = require('./modules/logger').init();
-const init    = require('./modules/init');
+const log = require('./modules/logger').init();
+const init = require('./modules/init');
 const _auth = require('./modules/webrequest/http-auth');
 
 // Keep a global reference of the window object, if you don't, the window will
@@ -35,38 +33,37 @@ let tray;
 // initialization and is ready to create browser windows.
 // Some APIs can only be used after this event occurs.
 app.on('ready', () => {
-
-  log.info('app ready')
+  log.info('app ready');
   log.debug('argv', process.argv);
   log.debug('options', options);
-  
+
   // initialize the authentication filter
   _auth.init();
-  
+
   initMainWindow();
   init.start(mainWindow);
 });
 
 // Quit when all windows are closed.
-app.on('window-all-closed', function () {
+app.on('window-all-closed', function() {
   // On OS X it is common for applications and their menu bar
   // to stay active until the user quits explicitly with Cmd + Q
   if (process.platform !== 'darwin') {
-    app.quit()
+    app.quit();
   }
 });
 
-app.on('activate', function () {
+app.on('activate', function() {
   // On OS X it's common to re-create a window in the app when the
   // dock icon is clicked and there are no other windows open.
   if (mainWindow === null) {
-    initMainWindow()
+    initMainWindow();
   }
 });
 
 // In this file you can include the rest of your app's specific main process
 // code. You can also put them in separate files and require them here.
-app.on('browser-window-created', function (e, window) {
+app.on('browser-window-created', function(e, window) {
   window.setMenu(null);
 });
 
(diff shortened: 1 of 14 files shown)


zaSmilingIdiot:
Additional changes added:

* A default wallet is created when the app is installed, but because it is not initialized, the create/restore wallet process is initiated. Successfully creating/restoring a wallet would create/restore a new wallet, leaving the previous wallet.dat unitialized, meaning that on subsequent launches of the application, the same create/restore process would be initiated as the default wallet was still in its default creation state. 
This has now been fixed: initial launching oft he app now correctly first sets the default wallet.

* If the user closes the app prior to correctly initializing a wallet (properly creating/validating recovery, etc), but after a wallet name has been provided (NB! a wallet file is created immediately after a valid wallet name as been provided and the user has navigated to the next step), a new wallet was being created but would then fail when selecting it from the multiwallet menubar on subsequent app launch.
This has now been resolved: selecting an existing wallet that has not been properly initialized now completes the initialization process for that wallet (the user still has the choice of whether to restore an existing wallet to the wallet or to continue creating a new one).

* A "Cancel" button has been added to the initial page of the create/restore wallet process, allowing the user to navigate back to the main application module.
NB! The default wallet is selected as being the active wallet. This can be addressed, but I thought it to be more prevalent to do when actioning the starting up of the application with a default wallet active (similar problem scope).

src/app/installer/create-wallet/create-wallet.component.ts
```diff
@@ -83,8 +95,13 @@ export class CreateWalletComponent implements OnDestroy {
   }
 
   initialize(type: number): void {
+    const wName: string = this.name;
     this.reset();
 
+    if (this.isExistingWallet) {
```
kewde: seems a bit pointless?

src/app/installer/create-wallet/create-wallet.component.ts
```diff
@@ -83,8 +95,13 @@ export class CreateWalletComponent implements OnDestroy {
   }
 
   initialize(type: number): void {
+    const wName: string = this.name;
```
kewde: combined with this?

src/app/multiwallet/multiwallet.service.ts
```diff
@@ -34,28 +33,38 @@ export class MultiwalletService implements OnDestroy {
...
     // and load the wallets in _list.
-    this.timer.takeWhile(() => !this.destroyed)
-      .subscribe(() => {
-        this._http.get('wallets://lists')
-        .map((response: any) => response.wallets)
-        .do((wallets) => wallets.forEach(w => w.fakename = w.name.replace('wallet_', '')))
-        .subscribe((wallets: IWallet[]) => this._list.next(wallets));
-      })
+    this.timer.takeWhile(() => !this.destroyed).subscribe(() => {
+      this.rpc.call('listwallets').subscribe(
```
kewde: rm the electron backend code if you're going this route and remove the chokadir dependency.

kewde:
Awesome first PR :+1:

src/app/installer/create-wallet/create-wallet.component.ts
kewde: I see now, I don't know how I skipped over that line.
Perhaps this.reset() can be called in ngOnInit, such that it doesn't overwrite it at all.
AFAIK it only sets it to the default values.

zaSmilingIdiot: `this.reset()` is called in the constructor, which could be moved to ngOnInit as well. `initialize()` however, is also used as the click handler for both the 'restore' and 'create' buttons, and so we likely want to be calling `this.reset()` whenstarting either the create or restore wallet process. Particularly if the user has already reached the first step for a new wallet, prinput a name and hit the 'back' button. In which case we need to be cleaning up things like error messages, the wallet name, etc. So `this.reset()` should probably be used here as well. Which then necessitates the need to keep the wallet name (only) if the wallet is an existing wallet.

main.js
```diff
@@ -86,35 +81,38 @@ function initMainWindow() {
...
 
     webPreferences: {
       webviewTag: false,
       nodeIntegration: false,
       sandbox: true,
       contextIsolation: false,
-      preload: path.join(__dirname, 'preload.js')
-    },
+      preload: path.join(__dirname, 'preload.js'),
+      webSecurity: !options.dev
```
zaSmilingIdiot: This line is the only change in this file (the remainder are formatting changes).

modules/multiwallet.js
```diff
@@ -19,34 +20,35 @@ let wallets = [
...
+  wallets.push({ name: wallet });
 }
 
-
 function init() {
   const particlDataDir = cookie.getParticlDatadirPath();
 
   // watch the folders
-  const glob = join(particlDataDir, 'wallet_*');
+  const glob = join(particlDataDir, 'wallets', 'wallet_*');
```
zaSmilingIdiot: The significant change in this file. The wallets are found in a dir 'wallets' inside of the particl data dir, so adding this to the path is necessary in order to find the wallets now.

zaSmilingIdiot:
The last commit fixes up the issue of the the multiwallets not correctly loading when selected (reporting a `Requested wallet does not exist or is not loaded` error).

modules/multiwallet.js
kewde: Checking up on this, new behavior.

PR #1121, opened 2018-08-18
addition filter functionality and hide unused stuff

vikas-cis:


src/app/core/market/api/bid/bidCollection.model.ts
```diff
@@ -29,9 +29,23 @@ export class BidCollection {
   }
 
   get filterOrders(): Bid[] {
+    // @TODO additionalFilter stuff should be remove once it handles via backend
+    if (this.additionalFilter) {
+      if ((this.additionalFilter.hideCompleted && this.additionalFilter.requiredAttention) ||
```
zaSmilingIdiot: This condition is a little confusing: Its completely determined by the value of `this.additionalFilter.requiredAttention`, 

Thus:
`if ((this.additionalFilter.hideCompleted && this.additionalFilter.requiredAttention) || this.additionalFilter.requiredAttention) {`

can be replaced entirely with the same logical truthness as 
`if ( this.additionalFilter.requiredAttention ) {`.

And I'm not sure that that is the intended behaviour...

vikas-cis: @zaSmilingIdiot there is reason behind as we have scenario when both get checked so on the basis of that we can filter the orders if you go through the files we have few more conditions that filter accordingly. I hope this make sense for you 

Thanks

vikas-cis:
@zaSmilingIdiot we handling those additional filters from frontend rather querying from api

src/app/core/market/api/bid/bidCollection.model.ts
zaSmilingIdiot: I think you might still want to review your logical condition though... the truthiness of it is equivalent to the truthiness of `this.additionalFilter.requiredAttention` alone, and `this.additionalFilter.hideCompleted` is not currently being considered in the conditional statement. I understand that it probably should be, which is why the condition as it currently is, is wrong.

Issue #582, opened 2018-01-04
Finalize GUI Settings page 


HenkSwardt:
We need to wrap up the development on the GUI Settings page. Please can we discuss and finalize specification on this issue.

See below screen of design.

![particl_wallet_settings](https://user-images.githubusercontent.com/28799964/34556584-8aefe486-f13f-11e7-8383-98db17094a01.jpg)

allienworks:
Is Settings going to be on a separate page or as a part of Console modal (someone mentioned that earlier)?
What will be included in Settings?

I can start working on it as soon as I have the info.

HenkSwardt:
Hi @allienworks 

This is currently in the GUI but hidden. I believe it has most of what is needed but was done in Angular2. We will need to convert the page to Material first.

HenkSwardt:
@anandsinghparihar will start with the refactoring to Material of the settings page for us. Once this has been completed we will investigate and list all outstanding features on the settings page.

anandsinghparihar:
@HenkSwardt Setting page front-end related work has been done now need any backend developer to complete that that API related stuff (get and set setting), after completing API I'll refactor front-end like all pagination would be dynamically set by settings.

zaSmilingIdiot:
An internal task exists for this.
Given that multiwallets are now implemented in the v2.x implementations, there's a separation of "settings" related information: those that apply to the application as a whole (eg: core related settings), and those that apply specifically to a wallet (eg: whether the market is enabled on it or not).

I believe we need to perhaps re-address the design of this "settings" page to be able to accommodate the distinction. Perhaps even have 2 separate "settings": 
- system configuration
- wallet settings

Thoughts?

AllienWorks:
@zaSmilingIdiot that can be done, no problem. Do we have a overview of all the things that changed between the original PR state and today? I'll be happy to update it (we really need the Settings in place).

zaSmilingIdiot:
No, but I think there has been significant updates. I'm not too concerned with the development work of it -> I was re-working significant chunks of that implementation code at the time anyway. Part of that can be re-used though.

I think the main concern here is separating out what is actually significant in terms of the various settings. For example, indications of whether the side menus start open or closed doesn't seem to be a real concern, and we can add those in if there is a real request for that. More importantly, for example, is adding something like importing wallets via their .dat file, or deleting a wallet, for example (former is likely a "global" setting with the latter likely being a wallet specific one). 

[...]

PR #1122, opened 2018-08-22
GUI Settings (market + core)

cis-shubham-t:


AllienWorks:
Settings GFX preview: https://imgur.com/a/Z40rBNj

So far, I'm done. When you do your thing @cis-shubham-t, I will do the final tweaks afterwards ;)

cis-shubham-t:
Awesome @AllienWorks 
Thanks for the update :)

vikas-cis:
@zaSmilingIdiot @HenkSwardt please find the below sheet where everything has been added related to setting.

https://docs.google.com/spreadsheets/d/1Dit0CCHMkij7RsAV3ERMc1c8W8nQAHMd-OYhsFWgJJY/edit#gid=0

Thanks

AllienWorks:
We have some outgoing links in the GUI (regarding various "read more" links + all of the links in Help section) - can we whitelist those? Currently the links are not doing anything (worked before), I suspect it's some kind of security measure..

**Links to be whitelisted:**
- https://particl.wiki/ (whole domain)
- https://particl.io/ (whole domain or just https://particl.io/user-guides)
- https://riot.im/app/#/group/+particl:matrix.org
- https://discordapp.com/invite/2tVJaZ9
- https://t.me/particlproject
- https://particl.news/
- https://twitter.com/particlProject
- https://www.facebook.com/ParticlProject/
- https://www.reddit.com/r/Particl/
- https://plus.google.com/+Particl
- https://mastodon.social/@particl
- https://www.youtube.com/c/Particl
- https://bitcointalk.org/index.php?topic=1835782.0
- https://github.com/particl
- https://www.transifex.com/particl/particl-website/

src/app/wallet/settings/settings.component.html
```diff
@@ -444,7 +444,7 @@ <h4 class="option">Map port using UPnP</h4>
 
           <div class="item">
             <h4 class="option">Port forwarding</h4>
-            <p class="desc">Alternatively, you can configure your router to forward port <code>51738</code></p>
+            <p class="desc">Alternatively, you can configure your router to forward port<!-- <code>51738</code> --></p>
```
AllienWorks, line 447: Why did you hide this? How are users supposed to port forward, when they don't know the port? (The port doesn't change.)

zaSmilingIdiot: The port does change, depending on whether you are on testnet or mainnet :)
Also, while its not likely to be used outside of mainnet/testnet, the daemon can be configured with the 
`-port=<>`
flag, which does change the port value. This is also possible to be configured via the Desktop by passing it in as a flag when starting the Desktop client (or the daemon might have been configured via a particl.conf file containing the value).

AllienWorks:
Closes #582

zaSmilingIdiot:
The GUI here assumes that it is the source of truth for a variety of settings. 
This however, is not the general case: the particl.conf file can be created by a user individually, different datadir's for the deamon may be used, which allow for different daemon configuration, etc. Settings that relate to particl.conf changes, as well as those that read from the daemon need to be read in at the time of changing settings to ensure that the GUI has the latest values of those settings, rather than storing them and assuming that the stored values are the used values.

I've moved this to paused, as it is not priority at the moment.

vikas-cis-zz:
Hye @zaSmilingIdiot iirc the settings did change the conf setting on the time of changing it from gui but yeah it will not restart the daemon at the same time, the conf files changes will be applicable once it has been restarted

Thanks

zaSmilingIdiot:
True, but here are the more pressing concerns with the changes here:

1. Currently, writing to the particl.conf file overwrites whatever is in an existing particl.conf file. Since a user may have a custom particl.conf file, the GUI should not be overwriting this (deleting all other configuration specified in the file), but rather modifying any values if they are different, or otherwise creating them alongside the existing config.

2. Any configuration written to the particl.conf file is also currently stored in the GUI as well (along with the other settings the GUI allows to be configured). Where this is a problem is that the particl.conf file is never read back, so the GUI assumes that the changes it has stored itself (in localstorage) must be the correct settings. But this does not cater for the fact that the user may edit the particl.conf file themselves, or even allow for a different particl.conf file to be used. So when the GUI displays the Core settings to the user, the settings it displays is not necessarily what is currently in the configuration.

3. Extending point 3 above, there are a couple of settings that require interaction with the daemon. Again, since the user may set these values without using the GUI, we should be taking the action that when the user opens the settings page, the GUI query's the current values from the daemon and uses those as the current displayed values. There is no need to store those specific values.

I've already done some work in modifying the changes to fix up the above issues. Again, I'm just marking this as paused for now as other changes have priority at this point. I'll get back to working on it in the near future.

PR #1129, opened 2018-09-03
Moves daemon updater indication to loading screen.

zaSmilingIdiot:
Moves updater.service into loading.component.
Does away with the modal that pops up to indicate the downloading of a new daemon, instead displaying the downloading progress text inline in the loading page.


modules/multiwallet.js
```diff
@@ -27,7 +27,7 @@ function init() {
   const particlDataDir = cookie.getParticlDatadirPath();
 
   // watch the folders
-  const glob = join(particlDataDir, 'wallet_*');
+  const glob = join(particlDataDir, 'wallets', 'wallet_*');
```
kewde: I recall that this didn't work for me, I think I removed this once before.
I tested this with the 0.17 daemon

zaSmilingIdiot: I just updated this to look for wallets in both the default location and the additional 'wallets' dir.

PR #1166, opened 2018-09-27
Ring size option added in advance options.

cis-shubham-t:
- Added the ring size field in advance option.
- Update progress bar on the basis of ring size value.


AllienWorks:
I have a few issues with this.
1. this option should be visible only when Anon option is selected
2. why to hide this in Advanced options, when we have the slider related to RingCT already? (best to group stuff based on relevance)

screenshot from 2018-10-01 13-14-03

IMO best would be to:
- get rid of the current RingCT option in Adv. options
- in Anon's "Select Privacy level":
  - rename it to "Select Privacy level (Ring signatures)"
  - replace the progress bar by [formatted slider](https://material.angular.io/components/slider/overview#formatting-the-thumb-label) with labels for number of Ring sigs
  - keep "low - normal - high" labels for ordinary users (who has no idea about what Ring sigs are)

AllienWorks:
@cis-shubham-t I've prepared the code for your implementation. Did the changes I commented above:
- moved the RingCT settings from the right box to left sidebar
- removed `mat-progressbar` and added `mat-slider` instead
- added `Auto privacy` button for automatically setting the recommended privacy level (= medium?)
- all this is hidden until user clicks on `Advanced options` button (ordinary users don't really need to see it, default privacy would be used and it won't confuse them)

screenshot from 2018-10-02 15-52-54

### TODO

- `mat-slider` needs to be connected to backend

cis-shubham-t:
Thanks for updating me @AllienWorks,
I'll continue on the same.

src/app/wallet/wallet/send/send.component.ts
```diff
@@ -16,27 +15,21 @@ import { AddressLookUpCopy } from '../models/address-look-up-copy';
...
+} from 'app/modals/send-confirmation-modal/send-confirmation-modal.component';
 
 @Component({
   selector: 'app-send',
   templateUrl: './send.component.html',
   // TODO merge / globalize styles
-  styleUrls: ['./send.component.scss', '../../settings/settings.component.scss']
+  styleUrls: [
+    './send.component.scss',
+    '../../settings/settings.component.scss'
```
AllienWorks: Why is Settings SCSS included here? There's no reason for it AFAIK, only creates mess with style inheritance..

cis-shubham-t: Its already included in the market branch @AllienWorks.
here: https://github.com/particl/particl-desktop/blob/market/src/app/wallet/wallet/send/send.component.ts#L25.

Can we remove the same from here?

AllienWorks: The Settings styles were also included in Receive if I'm not mistaken, it was doing a huge styling chaos there (took me a while before I figured it out). So yes, remove it please. If any GFX issues arise, I will fix them. But we shouldn't definitely share component styles across multiple ones, thanks.

cis-shubham-t: Thanks, @AllienWorks for confirmation, I'll remove it shortly.

zaSmilingIdiot (approved):

zaSmilingIdiot:
References JIRA item https://particl.atlassian.net/browse/PD-427

(Updating of the adjustment of the ring size value for Anon send transactions)

Issue #1154, opened 2018-09-19
Another Minor Styling Issue with Proposals


cryptoguard:
As displayed in this screenshot: 
<img width="678" alt="screen shot 2018-09-19 at 2 45 16 pm" src="https://user-images.githubusercontent.com/27031337/45774438-eca1d300-bc1a-11e8-9e28-0b98c0fc384d.png">

The colored bulletpoints for voting options seem to be clipped somehow. This is on the latest build (the one that allegedly broken with 0.17).

zaSmilingIdiot:
Has been fixed.

PR #1168, opened 2018-09-29
- Updated the proposal option related config settings

cis-shubham-t:


cis-shubham-t:
Fixes related to the issue: #1154 
proposal

PR #1174, opened 2018-10-04
Add listing fix

vikas-cis-zz:


vikas-cis-zz:
https://particl.atlassian.net/browse/PD-380

src/app/market/sell/add-item/add-item.component.html
```diff
@@ -150,11 +150,11 @@
           <input matInput placeholder="Quantity">
         </mat-form-field-->
         <mat-form-field fxFlex="50%" class="auto-width">
-          <input matInput min="0" placeholder="Domestic shipping price" formControlName="domesticShippingPrice" type="number">
+          <input matInput min="0" (keypress)="inputValidate($event)" placeholder="Domestic shipping price *" formControlName="domesticShippingPrice" type="number">
```
zaSmilingIdiot: This still leaves the option of copy/pasting values into the input. This can be mitigated by including a (paste) event handler (possibly pointing to the same method - this would require updating the validation method to include a check for the pasted data  (see comment with the example on the function itself).

src/app/market/sell/add-item/add-item.component.html
```diff
@@ -141,7 +141,7 @@
       <!-- TODO: add another input for fiat value and inter-connect the two: -->
       <mat-form-field class="full-width">
         <mat-icon matPrefix class="icon" fontSet="partIcon" fontIcon="part-cryptocurrency-particl"></mat-icon>
-        <input matInput min="0" placeholder="Price per item" formControlName="basePrice" type="number">
+        <input matInput min="0" (keypress)="inputValidate($event)" placeholder="Price per item *" formControlName="basePrice" type="number">
```
zaSmilingIdiot: This still leaves the option of copy/pasting values into the input. This can be mitigated by including a (paste) event handler (possibly pointing to the same method - this would require updating the validation method to include a check for the pasted data  (see comment with the example on the function itself).

src/app/market/sell/add-item/add-item.component.html
```diff
@@ -150,11 +150,11 @@
           <input matInput placeholder="Quantity">
         </mat-form-field-->
         <mat-form-field fxFlex="50%" class="auto-width">
-          <input matInput min="0" placeholder="Domestic shipping price" formControlName="domesticShippingPrice" type="number">
+          <input matInput min="0" (keypress)="inputValidate($event)" placeholder="Domestic shipping price *" formControlName="domesticShippingPrice" type="number">
           <span matSuffix>PART</span>
         </mat-form-field>
         <mat-form-field fxFlex="50%" class="auto-width">
-          <input matInput min="0" placeholder="International shipping price" formControlName="internationalShippingPrice"
+          <input matInput min="0" (keypress)="inputValidate($event)" placeholder="International shipping price *" formControlName="internationalShippingPrice"
```
zaSmilingIdiot: This still leaves the option of copy/pasting values into the input. This can be mitigated by including a (paste) event handler (possibly pointing to the same method - this would require updating the validation method to include a check for the pasted data  (see comment with the example on the function itself).

src/app/market/sell/add-item/add-item.component.ts
```diff
@@ -315,6 +315,14 @@ export class AddItemComponent implements OnInit, OnDestroy {
     return this.itemFormGroup.valid || this.snackbar.open('Invalid Listing');
   }
 
+  inputValidate(event: any) {
```
zaSmilingIdiot: If using the same method handler to cater for paste events, an example updated function might look something as follows:

```
inputValidate(event: any) {
    // Special character validation
    const keys = ['-', 'e', 'E', '+'];
    const pasted = String(event.clipboardData ? event.clipboardData.getData('text') : '' );
    if (keys.includes(event.key) || pasted.split('').find((c) =>  keys.includes(c))) {
      return false;
    }
  }
```

I would also suggest renaming this function to indicate that the validation is for numerical fields.

src/app/market/sell/add-item/add-item.component.ts
```diff
@@ -315,6 +315,14 @@ export class AddItemComponent implements OnInit, OnDestroy {
...
+    // Special character validation
+    const keys = ['-', 'e', 'E', '+']
```
zaSmilingIdiot: Could we maybe define this array outside of the function? As a property on the class, perhaps? Its trivial reasoning, but technically we re-create this array everytime the function is invoked, and so we are recreating the array on every key press. It really is trivial in this specific case, but in general for performance reasons, the array does not change between invocations and so it can be created once at the creation of the component.

src/app/market/sell/add-item/add-item.component.ts
vikas-cis-zz: @zaSmilingIdiot for me on mac clipboardData doesn't return any data not sure why :(

zaSmilingIdiot (approved):

PR #1178, opened 2018-10-06
added button to reset the form

vikas-cis-zz:
https://particl.atlassian.net/browse/PD-424

### DIFF
--- a/src/app/market/buy/checkout-process/checkout-process.component.html
+++ b/src/app/market/buy/checkout-process/checkout-process.component.html
@@ -138,6 +138,10 @@
 
   <!-- =========== SHIPPING DETAILS =========== -->
   <mat-step [completed]="cache.selectedIndex >=2" [stepControl]="shippingFormGroup">
+    <button mat-raised-button color="primary" [disabled]="!profile?.shippingAddresses || profile?.shippingAddresses.length === 0" (click)="clearForm()">
+      <mat-icon class="icon" fontSet="partIcon" fontIcon="part-next-single"></mat-icon>
+      New Shipping Profile
+    </button>
     <form [formGroup]="shippingFormGroup" class="shipping">
       <ng-template matStepLabel>Shipping details</ng-template>
       <div fxLayout="row" fxLayoutGap="30px">
--- a/src/app/market/buy/checkout-process/checkout-process.component.html
+++ b/src/app/market/buy/checkout-process/checkout-process.component.html
@@ -138,10 +138,7 @@
 
   <!-- =========== SHIPPING DETAILS =========== -->
   <mat-step [completed]="cache.selectedIndex >=2" [stepControl]="shippingFormGroup">
-    <button mat-raised-button color="primary" [disabled]="!profile?.shippingAddresses || profile?.shippingAddresses.length === 0 || !addressNotSelected" (click)="clearForm()">
-      <mat-icon class="icon" fontSet="partIcon" fontIcon="part-next-single"></mat-icon>
-      Clear Shipping Address form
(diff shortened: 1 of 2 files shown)


vikas-cis-zz:
@AllienWorks I have added the button which clear the form tho if you can have a look and feel free to do css changes if required :)

Thanks

src/app/market/buy/checkout-process/checkout-process.component.ts
```diff
@@ -218,6 +218,13 @@ export class CheckoutProcessComponent implements OnInit, OnDestroy {
     this.getCache();
   }
 
+  clearForm(): void {
```
zaSmilingIdiot, line 221: A couple of notes on this:
1) When this 'New Shipping Profile' button is clicked on, this does not clear the selected shipping profile dropdown. This is 
(a) confusing: it still shows that a shipping profile has been selected when the user is in fact creating a new one; 
(b) if the user changes their mind and decides they want to use that profile after all, they first have to select another profile (if there is one, which is another problem if there is only one saved profile at the moment) and then re-select the previously selected profile. This is because a change event is emitted for a select box when the chosen option value is different to its current value.

So the 'selectedAddress' field needs to be cleared as well.

(2) Is there any reason why the clear() method defined previously is not updated and used?

src/app/market/buy/checkout-process/checkout-process.component.ts
```diff
@@ -218,6 +218,13 @@ export class CheckoutProcessComponent implements OnInit, OnDestroy {
...
+    if (Object.keys(this.selectedAddress).length > 0) {
+      this.selectedCountry = undefined;
```
zaSmilingIdiot: Does this need to be explicitly cleared? Is it not cleared as part of clearing the form?

src/app/market/buy/checkout-process/checkout-process.component.html
```diff
@@ -138,6 +138,10 @@
 
   <!-- =========== SHIPPING DETAILS =========== -->
   <mat-step [completed]="cache.selectedIndex >=2" [stepControl]="shippingFormGroup">
+    <button mat-raised-button color="primary" [disabled]="!profile?.shippingAddresses || profile?.shippingAddresses.length === 0" (click)="clearForm()">
```
zaSmilingIdiot: I think the button should probably be disabled if no saved profile shipping profile has been selected. In other words, If a saved profile has not been selected, then this button should be disabled. 

Alternatively, change the button text to read something similar to 'Clear Shipping Address form', as that is essentially its purpose. In which case it would make sense to disable the button if the entire form is empty, and it would be better then to move the button to the form that it clears.

zaSmilingIdiot (changes requested):

src/app/market/buy/checkout-process/checkout-process.component.ts
vikas-cis-zz: no it requires explicitly clearing

AllienWorks:
While testing this, I see that the bug of "moved fields" appeared again:

shipping-bug

- my first name is in profile
- last name is in first name
- profile is in last name

Can you please check and fix this?

vikas-cis-zz:
@AllienWorks can you provide the steps that you are doing see this ?

THanks

AllienWorks:
@vikas-cis nevermind, probably something in db changed. I saved this profile a while ago, every branch is doing it as well. But when I create a new Shipping profile, it's all OK. Forget about it :)

zaSmilingIdiot (approved):

PR #1180, opened 2018-10-08
fix for selecting the default value

vikas-cis-zz:
https://particl.atlassian.net/browse/PD-434

### DIFF
--- a/src/app/core-ui/material/mat-otpgroup-select-search/mat-otpgroup-select-search.component.ts
+++ b/src/app/core-ui/material/mat-otpgroup-select-search/mat-otpgroup-select-search.component.ts
@@ -97,6 +97,10 @@ export class MatOtpGroupSelectSearchComponent implements OnInit, OnChanges {
   }
 
   onBlur($event: any): void {
+    if ($event.target.value === this.defaultOption) {
+      return;
+    }
+
     if (this._filterGroup($event.target.value, 'blur').length === 0) {
       this.textInput.nativeElement.value = '';
     }
--- a/src/app/core-ui/material/mat-select-search/mat-select-search.component.ts
+++ b/src/app/core-ui/material/mat-select-search/mat-select-search.component.ts
@@ -81,6 +81,10 @@ export class MatSelectSearchComponent implements OnInit, OnChanges {
   }
 
   onBlur($event: any): void {
+    if ($event.target.value === this.defaultOption) {
+      return;
+    }
+
     if (this._filter($event.target.value, 'blur').length === 0) {
       this.textInput.nativeElement.value = '';
     }


vikas-cis-zz:
@HenkSwardt @zaSmilingIdiot please have a look and let me know if the selecting behaviour looks good :)

Thanks

src/app/core-ui/material/mat-otpgroup-select-search/mat-otpgroup-select-search.component.ts
```diff
@@ -97,6 +97,10 @@ export class MatOtpGroupSelectSearchComponent implements OnInit, OnChanges {
   }
 
   onBlur($event: any): void {
+    if ($event.target.value === this.defaultOption) {
```
zaSmilingIdiot, line 100: Do we need this extra condition here? It would seem that the same effect could have been achieved by removing the content of the onBlur() method:
The code snippet
```
if (this._filterGroup($event.target.value, 'blur').length === 0) {
      this.textInput.nativeElement.value = '';
    }
```
appears to exist for the purpose of setting the input value to an empty string if no selection has been made. Which is what we are trying to undo. Removing this condition (without introducing any new code), would surely have the same effect?

src/app/core-ui/material/mat-select-search/mat-select-search.component.ts
```diff
@@ -81,6 +81,10 @@ export class MatSelectSearchComponent implements OnInit, OnChanges {
   }
 
   onBlur($event: any): void {
+    if ($event.target.value === this.defaultOption) {
```
zaSmilingIdiot, line 84: Do we need this extra condition here? It would seem that the same effect could have been achieved by removing the content of the onBlur() method:
The code snippet
```
    if (this._filter($event.target.value, 'blur').length === 0) {
      this.textInput.nativeElement.value = '';
    }
```
appears to exist for the purpose of setting the input value to an empty string if no selection has been made. Which is what we are trying to undo. Removing this condition (without introducing any new code), would surely have the same effect?

zaSmilingIdiot (changes requested):

src/app/core-ui/material/mat-select-search/mat-select-search.component.ts
vikas-cis-zz: yes we need as requested in the ticket PD-434 and we need to show the selected default option if user select it 

Thanks

src/app/core-ui/material/mat-otpgroup-select-search/mat-otpgroup-select-search.component.ts
zaSmilingIdiot: As mentioned by @vikas-cis the existing condition is necessary to ensure that the input text correctly matches an available option. If the user types in text that matches no available options, then the invalid input needs to be cleared.
This clearing of the input for filter options is the basis of PD-437

src/app/core-ui/material/mat-select-search/mat-select-search.component.ts
zaSmilingIdiot: As mentioned by @vikas-cis the existing condition is necessary to ensure that the input text correctly matches an available option. If the user types in text that matches no available options, then the invalid input needs to be cleared.
This clearing of the input for filter options is the basis of PD-437

zaSmilingIdiot (approved):

PR #1182, opened 2018-10-09
filter fix as suggested in PD-437

vikas-cis-zz:
https://particl.atlassian.net/browse/PD-437


vikas-cis-zz:
@zaSmilingIdiot I have checked the scenarios according to me for all the filter on all pages where it has been used , please let me know if there is anything that we still missing 

Thanks

zaSmilingIdiot (changes requested):
A few notes on this:

1. There is still an issue with the dropdown select elements in the 'Add Listing' page, demonstrated by the following sequence:
    - if some text is input into the box, the options are filtered correctly according to the typed-in text; 
    - if the user then clicks away from the selection, the input is cleared; 
    - when the user clicks back into the input, the listed options are still filtered as per the previous typed-in text (which was cleared).
    - The correct behaviour in this case is to clear the input but also to ensure that the filtered options are reset. 
2. The use case for leaving the current value for the filter options generally works as mentioned in the PD-437 issue. However, following on from (1) above, we should be able to reset the input to display the currently selected option, as well as ensuring that the filtered list of options are reset as well.
3. If using a 'pageFrom' type of variable to determine whether to clear values or not, it might be better to make use of a boolean type of value: either the input is cleared or it is not - there are only 2 options available.

vikas-cis-zz:
@zaSmilingIdiot please have a look and let me know your valuable idea and suggestions 

thanks

vikas-cis-zz:
@zaSmilingIdiot I have fixed another scenario, please have a look and let me know 

thanks

zaSmilingIdiot (approved):
Use of settimeout is not the best way to achieve this, but it is the current best option available.

These generic components will likely be replaced in the very near future so this is sufficient as an interim solution.

PR #1185, opened 2018-10-10
Fixes bug in Send Anon to Public transfer

zaSmilingIdiot:
This fixes 2 issues:
1. In Wallet > Send, transferring funds from Anon to Public resulted in an error due to an out of range value used for the inputs_per_sig field in the sendtypeto rpc command (fixed by now using the deamon default value),
2. After cancelling the modal asking for confirmation of a balance transfer, the user would need to manually clear the 'receiver address' input value in the Advanced section (automatically populated in a balance transfer), before being able to try another transfer (fixed by automatically clearing the receiver address value on cancel).

Also cleans up some unnecessary validation.


zaSmilingIdiot:
Has reference to https://particl.atlassian.net/browse/PD-430

PR #1187, opened 2018-10-12
Proposal content fix

cis-shubham-t:
I saw one issue on the proposal of the description content.
before fix: 
before

I've fixed the same with the help of @AllienWorks.

After fix:

after


cis-shubham-t:
https://particl.atlassian.net/projects/PD/issues/PD-441

zaSmilingIdiot (approved):

PR #1189, opened 2018-10-15
Fixes bug with flagging of listing items

zaSmilingIdiot:
If a user has not voted on an flagged item, check if the submitter of the proposal was the current user profile, and display the correct message.


zaSmilingIdiot:
Closing direct repo PR in favour of submission via external repo (@vikas-cis will potentially be adding some changes to this, and potentially needs push access).

Issue #1155, opened 2018-09-20
Category Dropdown Menu Styling Enhancement Proposal


cryptoguard:
Right now, when selecting a category this is what the dropdown menu looks like: 
<img width="294" alt="screen shot 2018-09-19 at 9 56 27 pm" src="https://user-images.githubusercontent.com/27031337/45791323-2a245180-bc57-11e8-9aae-ec129b9c621f.png">

I was thinking perhaps it could be easier to navigate through the menu if all categories were collapsed, and only expanded once you click on it. Not sure if it is possible within a dropdown menu though...

So for example, in the screenshot above, instead of having (in order) the following when opening the dropdown menu: 
- Particl
- - Free Swag
- High Values
- - Business / Corporate
etc

You would have: 
- Particl
- High Values
etc

and only once you click on High Values would "Business / Corporate", as well as the other sub-categories for High Values, would show up. This would make it easier to choose the right category and make navigating through the options less of a hassle.

kasprrr:
To add to this, whenever a user presses the letter K, U or whatever on its keyboard, categories/countries starting with that specific letter should come up automatically (fairly standard drop-down behavior). Right now this is not the case

HenkSwardt:
Hi @kasprrr 

Can you please provide me with your versions of the GUI as displayed at the bottom left corner of the Application as well as operating system details.

Reason I'm asking, we added this behavior to allow for hot searches a while back and it is working on my side. Please see below:

Countries:

![image](https://user-images.githubusercontent.com/28799964/45939878-02761600-bfd6-11e8-88c0-7c4de362b6b8.png)


Categories:

![image](https://user-images.githubusercontent.com/28799964/45939898-22a5d500-bfd6-11e8-82d1-012107207ab1.png)

zaSmilingIdiot:
@kasprrr Do you have any feedback for us on this?

zaSmilingIdiot:
@cryptoguard Thanks for the suggestion. We are currently looking into completing the suggested menu behaviour.

HenkSwardt:
For now we have added a tooltip on hover to display any characters that might not fit into the display space, 
Please see: 
https://particl.atlassian.net/browse/PD-363
https://github.com/particl/particl-desktop/pull/1176

We also added a task to further add the ability to expand and collapse the main categories, giving you a more detailed view of all main categories. 
Please see: 
https://particl.atlassian.net/browse/PD-436

kasprrr:
[...]

PR #1202, opened 2018-10-17
Correct country autocomplete option filtering

zaSmilingIdiot:
Filtered options for country selection are now matched to the input text by the start of the country name, rather than whether the country option contains the input text.


zaSmilingIdiot:
Has reference to: https://github.com/particl/particl-desktop/issues/1155#issuecomment-428721077

PR #1204, opened 2018-10-17
button will be disable if user renter the unknown value

vikas-cis-zz:
https://particl.atlassian.net/browse/PD-462


vikas-cis-zz:
@zaSmilingIdiot please let me know if we miss any scenario in that or it looks good now as per as issue suggested 

Thanks

zaSmilingIdiot (approved):

PR #1206, opened 2018-10-18
prevent duplicate address generated by updating the existing profile …

vikas-cis-zz:
…if it is selected

https://particl.atlassian.net/browse/PD-435


vikas-cis-zz:
@zaSmilingIdiot please have a look and let me know if that looks fine and if there is any suggestion 

thanks

zaSmilingIdiot (approved):

PR #1210, opened 2018-10-20
scroll fix

vikas-cis-zz:


vikas-cis-zz:
@zaSmilingIdiot I have reproduce the button doesn't enable in the case when there is no scroll bar so may be that is the case with other users for the big resolution size doesn't have scrollbar available on it so on the basis of that I have fixed that :)

Thanks

AllienWorks:
Just tested. The checkbox isn't implemented, but the issue with always disabled button is fixed.

I had to simulate this via removing half of the text as well, but it works. If scrollbar isn't present, button is available right away.

I think this is good enough. After all, I presume the final Terms will be looong so much, that nobody's screen could display it at once :)

zaSmilingIdiot (approved):

PR #1212, opened 2018-10-22
Fix pool zapping & updating per branch

kewde:


zaSmilingIdiot:
Looks like whitespace issues are causing some linting problems, which is why the build failed :smiling_imp:

zaSmilingIdiot (approved):

PR #1218, opened 2018-10-25
converting the error(if error comes in inconsistent format) to string

vikas-cis-zz:


zaSmilingIdiot (changes requested):
The changes included so far do appear to solve the problem in this particular case. However, there are a few other scenarios where the marketplace error string is wrapped inside of an extra JS object. To solve the more general problem, I would rather implement the necessary error string extraction in _MarketService#call()_

vikas-cis-zz:
@zaSmilingIdiot I agreed but still to avoid any case of giving that error so that we can atleast check for string :)

zaSmilingIdiot (approved):

Issue #1198, opened 2018-10-16
UI/UX Improvement Suggestion on Shipping Profiles


cryptoguard:
When buying an item for the first time, the user doesn't have a shipping profile registered (unless that is possible in the future with the Settings page). 

Here's what it looks like: 
<img width="449" alt="screen shot 2018-10-16 at 6 19 24 pm" src="https://user-images.githubusercontent.com/27031337/47051021-77b9ae80-d170-11e8-9558-0fe7f7e0449a.png">

When clicking on the dropdown menu, there is no action, which is kinda going against the user expectation of at least an empty dropdown menu. Perhaps, instead of having the menu being inactive, it could drop down with a single choice that says something like "None" or "Please save a shipping profile first". 

Perhaps there are better ways to do it as well. This is a minor detail but something I noticed as odd/unsatisfying.

zaSmilingIdiot:
Thanks for the suggestion :)

This may be better served if we can disable the shipping profile dropdown if it contains no saved options.

HenkSwardt:
This issue and further related issues will be addressed as part of this issue logged:

https://particl.atlassian.net/browse/PD-469

PR #1229, opened 2018-11-02
fix the back to update the shipping address title

vikas-cis-zz:
https://particl.atlassian.net/browse/PD-469

### DIFF
--- a/src/app/market/buy/checkout-process/checkout-process.component.ts
+++ b/src/app/market/buy/checkout-process/checkout-process.component.ts
@@ -106,6 +106,7 @@ export class CheckoutProcessComponent implements OnInit, OnDestroy {
 
   ngOnDestroy() {
     this.destroyed = true;
+    this.clear();
     this.storeCache();
   }
 
@@ -216,8 +217,9 @@ export class CheckoutProcessComponent implements OnInit, OnDestroy {
 
   select(address: Address, isNewProfile?: boolean) {
     this.log.d('Selecting address with id: ' + address.id);
+    // check for the new profile and then add.
     if (isNewProfile) {
-      this.profile.shippingAddresses = [address];
+      this.profile.shippingAddresses.push(address);
     }
     this.selectedAddress = address;
     this.shippingFormGroup.value.id = address.id;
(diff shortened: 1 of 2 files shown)


src/app/market/buy/checkout-process/checkout-process.component.ts
```diff
@@ -214,8 +214,11 @@ export class CheckoutProcessComponent implements OnInit, OnDestroy {
     })
   }
 
-  select(address: Address) {
+  select(address: Address, isNewProfile?: boolean) {
     this.log.d('Selecting address with id: ' + address.id);
+    if (isNewProfile) {
+      this.profile.shippingAddresses = [address];
```
zaSmilingIdiot: When navigating from step 3 to 2 int he checkout process, the shipping profile is now correctly selected, if one was previously selected. However, it is the only option in the shipping profile dropdown, which I suspect is a consequence of this specific line of code.

zaSmilingIdiot (changes requested):
Shipping addresses appear to be obtained only once, during the initialisation of the form. Specifically, when the following line is executed:
`this.getProfile();`

The shipping addresses should probably be updated and a shipping profile address re-selected if one was previously selected, when the user navigates back from step 3 to step 2 of the checkout process.

This PR also seems to be missing the change to return to step 1 of the checkout process when the user navigates to a different page and back.

vikas-cis-zz:
@zaSmilingIdiot it is still in progress so obviously the second part hasn't been pushed yet :P :D 

Thanks

vikas-cis-zz:
@zaSmilingIdiot yeah now you can check the branch as ready for review :)

Thanks

zaSmilingIdiot:
@vikas-cis Pushing the newly created shipping profile into the array of shipping profiles does not consider a number of scenarios, such as:
- if the shipping profile is already in the array (if the user has navigated back and forth between steps 2 and 3 multiple times),
- if the the address is being updated instead of a new address being created.

There are a couple more scenarios to consider.
Nevertheless, I've updated the code to reflect what I think is the correct way to handle all the scenarios I can think of. Essential is that:
a. if the user has selected a saved address AND modified the address but NOT requested the address to be saved: ensure that the previously selected address is populated in the dropdown and the modified form values are displayed with their modifications when returning to step 2;
b. if the user has selected a saved address AND modified the address but HAS requested the address to be saved: has the same result as (a), but with the address having been updated (this includes the saved address name)
c. if the user has NOT selected a saved address: the same results from (b) should apply but the dropdown should be empty when returning to step 2.

For sanity checking, can you please verify that this holds and that nothing else has broken?

Thanks.

vikas-cis-zz:
@zaSmilingIdiot yeah I gotch you :)

Thanks

vikas-cis-zz:
@zaSmilingIdiot  yeah I did look and seems working well to me 👍 

Thanks

zaSmilingIdiot (approved):

zaSmilingIdiot:
Fixes https://github.com/particl/particl-desktop/issues/1198

Issue #1243, opened 2018-11-09
Hour Stamps on Overview


cryptoguard:
Currently, the Overview page does not display the times at which orders/transactions/proposals are created. See these two screenshots: 

<img width="650" alt="screen shot 2018-11-09 at 1 53 21 pm" src="https://user-images.githubusercontent.com/27031337/48282169-e2d06b00-e426-11e8-910e-186568b70439.png">

<img width="644" alt="screen shot 2018-11-09 at 1 53 29 pm" src="https://user-images.githubusercontent.com/27031337/48282172-e8c64c00-e426-11e8-81bb-124460d4dd2e.png">

You can clearly see all events happened on the 9th of November, however, it is impossible to tell at what time each of these events happened. This is an important information to have at times. The only way right now to know the time is to click on the TX ID and look at the time on the block explorer (which is not ideal from a user-friendliness and accessibility standpoint).

PR #1245, opened 2018-11-12
Adds timestamps to the history and overview sections' transaction tables.

zaSmilingIdiot:
Also:
* fixes a minor issue with errors when attempting to list a broken proposal.


zaSmilingIdiot:
Fixes https://github.com/particl/particl-desktop/issues/1243

PR #1258, opened 2018-11-14
sync modal should run in the background and createwallet will only op…

vikas-cis-zz:
…en when terms are accepted

https://particl.atlassian.net/browse/PD-472


vikas-cis-zz:
@zaSmilingIdiot please have a look and let me know if this is all we need to do :)

Thanks

zaSmilingIdiot (approved):

PR #1262, opened 2018-11-15
Listing item information update (Edit un-published listing)

cis-shubham-t:
- https://particl.atlassian.net/projects/PD/issues/PD-470


src/app/market/sell/add-item/add-item.component.html
```diff
@@ -63,7 +63,7 @@
         <div class="title">Listing expiration &amp; fees</div>
         <div class="row" fxLayout fxLayoutAlign="center center">
           <mat-form-field class="full-width" fxFlex="1 1 100%">
-            <mat-select placeholder="Listing expiration" [(ngModel)]="expiration" (change)="loadTransactionFee()">
+            <mat-select placeholder="Listing expiration" [disabled]="isExistingTemplate()" [(ngModel)]="expiration" (change)="loadTransactionFee()">
```
zaSmilingIdiot: When editing a saved template, the 'Listing & Expiration fees' select is not editable because of this condition. However, if the user wishes to now Save and Publish the listing, they are not able to edit this field. 

A better option may be to not disable editing of this field. Instead, perhaps an option would be to display a large notification below this input if the template has an 'unpublished status', informing the user that they may need to edit this field as it is not saved as part of the template.

zaSmilingIdiot (changes requested):
Can the payment.service.ts and shipping.service.ts functions not be included in the template service rather than creating what is essentially single function service values? This keeps the template information together, and saves importing additional services unnecessarily.

Yes, I realize that the escrow, information, etc services all do this, but its unnecessary unless the service provides functionality that is useful (and is used) on its own. In this case, it really is not used separately from the remainder oft he template functionality, and so we can keep the functionality together.

cis-shubham-t:
Hey @zaSmilingIdiot 
As per our discussion, I added the 'select expiry time' option in the expiry dropdown by the same we can assure that user selected the expiry time as we are not providing the expiry time edit feature for now while the edit listing.

cis-shubham-t:
hey @AllienWorks 
I've moved the listing expiry drop-down in the separate modal now. There is some UI fixes and changes required, could you please have look over it and make the proper.
Thanks

AllienWorks:
There is a bug – Listings that are saved for later (not published immediately) and then Published later from Sell > Listings, don't trigger the expiration modal & Listings get published right away.

Plus I have some weird behaviour with my "Silver Pineapple" Listing:
- in Sell > Listings, I can see it as "Unpublished" (even if it's already published and visible on Market)
- it behaves like someone else's Listing (I can put it in my cart, even if I own it)
- tried to buy it and yes – I can buy/sell it to myself (Order shows up both in my Sell and Buy), but when Rejecting/Accepting the bid, I'm getting `Not your ListingItem` error

This Listing was published like I described in first paragraph (saved for later, published afterwards). Tried another time (with "Private $PARTyville Island"), but that one is behaving correctly for some reason.

---

@cis-shubham-t you had a typo in the component, so I renamed it, FYI.

AllienWorks:
Styling done, passing this back to you @cis-shubham-t

zaSmilingIdiot (approved):
Seems to work as expected now :)

Issue #1153, opened 2018-09-18
No Button to Clear Proposal Notifications


cryptoguard:
As a user, I might not want to vote on every single proposals being submitted and still want to clear the notification count: 

<img width="232" alt="screen shot 2018-09-18 at 4 46 57 pm" src="https://user-images.githubusercontent.com/27031337/45715588-95d8c280-bb62-11e8-92aa-fb4791dbef8b.png">

There should be a button or way to clear the notification count back to 0 without having to vote on every single proposals.

zaSmilingIdiot:
Thanks for the suggestion. This might be achieved by clearing the notification count when the proposals section is opened. The newest proposals are displayed first anyway, so is readily visible to the user when opening the page. 

Side note: we would then additionally need to ensure the notification count is not updated while the proposals page is currently open.

HenkSwardt:
This will be looked at in https://particl.atlassian.net/browse/PD-467

zaSmilingIdiot:
Has been implemented.

PR #1271, opened 2018-11-21
Clear proposal notification count.

cis-shubham-t:


src/app/core/market/proposals-notifier/proposals-notifications.service.ts
```diff
@@ -39,26 +44,44 @@ export class ProposalsNotificationsService implements OnDestroy {
...
+      });
+  }
+
+  // @TODO remove once functionality done from the MP side.
+  loadstoredProposals(): void {
+    this.storedProposals = JSON.parse(localStorage.getItem('proposals')) || [];
+  }
+
+  // @TODO remove once functionality done from the MP side.
+  storeProposals(): void {
```
zaSmilingIdiot: Is there a reason that we're storing in this method the actual proposals rather than just the last viewed timestamp value?

cis-shubham-t:
Issue #1153

src/app/core/market/proposals-notifier/proposals-notifications.service.ts
```diff
@@ -14,6 +14,7 @@ export class ProposalsNotificationsService implements OnDestroy {
 
   log: any = Log.create('order-status-notifier.service id:' + Math.floor((Math.random() * 1000) + 1));
   public proposals: Proposal[] = [];
+  public storedProposals: Proposal[] = [];
```
zaSmilingIdiot: Rather than store a list of proposals, it would be better to store a timestamp value, indicating the last time the proposals were viewed.

src/app/core/market/proposals-notifier/proposals-notifications.service.ts
```diff
@@ -39,26 +44,44 @@ export class ProposalsNotificationsService implements OnDestroy {
...
+          // get the count of the newly arrivied proposals.
+          this.proposalsCountRequiredVoteActions = this.getNewProposals(proposals, this.storedProposals).length
+        }
+        this.proposals = proposals;
+      });
+  }
+
+  // @TODO remove once functionality done from the MP side.
+  loadstoredProposals(): void {
+    this.storedProposals = JSON.parse(localStorage.getItem('proposals')) || [];
```
zaSmilingIdiot: This should read in a timestamp value that has been saved.

src/app/core/market/proposals-notifier/proposals-notifications.service.ts
```diff
@@ -39,26 +44,44 @@ export class ProposalsNotificationsService implements OnDestroy {
...
-
-      if (this.proposals.length !== proposals.length) {
-        this.checkProposalsRequiredVoteActions(proposals);
-      }
-      this.proposals = proposals;
-    });
+      .list(Date.now(), '*')
+      .take(1)
+      .subscribe((proposals: Proposal[]) => {
+        proposals.reverse();
```
zaSmilingIdiot: This shouldn't be necessary. If we know that the list of proposals is already ordered by date, then this line is unnecessary: either traverse the array forwards or backwards when performing the subsequent calculations. The proposals are not being displayed from this call, so adding the reversal is unnecessary overhead.

src/app/wallet/proposals/proposals.component.ts
```diff
@@ -171,6 +173,7 @@ export class ProposalsComponent implements OnInit, OnDestroy {
   }
 
   ngOnDestroy() {
+    this.proposalsNotificationsService.storeProposals();
```
zaSmilingIdiot: I don't think we need this method call. It can be changed rather into a method call to ensure that the timestamp is updated with the current value. Also possibly include a boolean value to indicate that the proposalNotificationService should start notifying the user of new proposals again.

src/app/wallet/proposals/proposals.component.ts
```diff
@@ -52,7 +53,8 @@ export class ProposalsComponent implements OnInit, OnDestroy {
   constructor(
     private router: Router,
     private peerService: PeerService,
-    private proposalsService: ProposalsService
+    private proposalsService: ProposalsService,
+    private proposalsNotificationsService: ProposalsNotificationsService
   ) { }
 
   ngOnInit() {
```
zaSmilingIdiot, line 63: Similar to ngOnDestory, we may need to update the proposalNotificationService with the current timestamp. This should clear the notification to the user, as the current timestamp would be greater than the that of the proposals.

Additionally, adding a boolean argument in may be necessary to indicate to proposalNotificationService to stop informing the user about new proposals, as the user would currently be on the proposals page. The opposite boolean value would then be passed in during the ngOnDestroy method to indicate that the proposalNotificationService should start notifying the user of new proposals again.

src/app/core/market/proposals-notifier/proposals-notifications.service.ts
```diff
@@ -14,6 +14,7 @@ export class ProposalsNotificationsService implements OnDestroy {
 
   log: any = Log.create('order-status-notifier.service id:' + Math.floor((Math.random() * 1000) + 1));
   public proposals: Proposal[] = [];
```
zaSmilingIdiot: Do we need to store proposals? Or should this not rather be a numerical value indicating the number of proposals that are new?

There doesn't seem to be much of a point in saving all of the proposals.

zaSmilingIdiot (changes requested):
Generally, I think that I've left a lot of comments which might seem confusing. The overall change changes I'm generally attempting to indicate are as follows:
- Make use of a last timestamp value (save a timestamp of when last the proposals page was previously opened);
- There is no need to store a reference/copy of any proposals themselves inside ProposalsNotificationsService  - by using the timestamp value as mentioned above, you simply need to check whether any new proposal have been created after the _last_viewed_ saved timestamp value - if so count how many there are and keep a copy of that specific number for use elsewhere (such as displaying to the user);

Additionally, there are unnecessary inclusions/references of proposals-notifications.service elsewhere in the app that can now be removed. Specifically, since we are no longer updating the notification count based on user votes, the references in proposal-details.component.ts can be removed as part of this PR.

src/app/core/market/proposals-notifier/proposals-notifications.service.ts
```diff
@@ -68,6 +71,9 @@ export class ProposalsNotificationsService implements OnDestroy {
             }
             if (needUpdating) {
               this.notifcationTimestamp = Date.now();
+
+              // restrict the notification for the unseen proposal at time time of GUI started.
+              localStorage.setItem(this.storageKeys.timestamp_notifcation, String(this.notifcationTimestamp));
```
zaSmilingIdiot, line 76: This line can be moved into the storeLastViewedProposalTimestamp() method (which can also be renamed if you feel up to it :) ). 
Rather keep the saving/writing of data in the same method, and leave this function to its purpose of data processing.

zaSmilingIdiot (changes requested):
Just one last thing to be resolved

src/app/core/market/proposals-notifier/proposals-notifications.service.ts
cis-shubham-t: @zaSmilingIdiot 
As I mentioned before in our conversation we need to update the timestamp_notificaiton, in that case as well if the user does not go to the proposal page.

If I put that code in the storeLastViewedProposalTimestamp()
then timestapm_notification only in that case when I go to the proposal page, as storeLastViewedProposalTimestamp() called in that case only with the help if viewingProposals().

So It's hard to maintain for first time notification (afterwords GUI started). if not updated the timestamp in the storage after notification pops up.

zaSmilingIdiot (approved):
All good :)

PR #1281, opened 2018-11-27
fixed updated time issue

EagleEye25:
fixed updated time that was the same as the original creation time on the orders tab


zaSmilingIdiot (changes requested):
Can you perhaps verify your yarn.lock file changes?
There are a couple of new additions in there that we definitely don't use (and probably need/want). For example, I found these:
* camelcase
* dockerfile-ast
* vscode-languageserver-types

vikas-cis-zz (changes requested):
@EagleEye25  not sure why you deleted the yarn lock file but I think it shouldn't iirc @zaSmilingIdiot  ?

zaSmilingIdiot (approved):

PR #1282, opened 2018-11-27
Create display expiration date

EagleEye25:


src/app/core/market/api/template/template.model.ts
```diff
@@ -175,6 +175,11 @@ export class Template {
     return this.object.FlaggedItem;
   }
 
+  get expiredAt(): any {
+    const expire = new DateFormatter(new Date(this.object.ListingItems[0].expiredAt)).dateFormatter(false).substr(0, 16);
```
zaSmilingIdiot: If this.object does not have ListingItems or this.object.ListingItems is an empty array, then this line will throw an exception.

zaSmilingIdiot (changes requested):
Requires a couple of checks that ensure existence before usage.

src/app/core/market/api/template/template.model.ts
EagleEye25: Completed

vikas-cis-zz:
@EagleEye25 didn't saw your last update :P it is much better :D :D nevermind 

Thanks

zaSmilingIdiot (approved):

PR #1284, opened 2018-11-28
Community Test Build 2018-11-28

zaSmilingIdiot:
Preparation for community test build.


zaSmilingIdiot:
Irrelevant build required now. Still requires an investigation/fix for fiona-builds bot, but for now the builds have made it to the community through alternative means.

PR #1304, opened 2018-12-06
Order: expand shipping area

### LINKED ISSUE #1303: Cannot See the Complete Address from Customers
Trying to process lapel pin orders, and cannot see the full addresses of customers in the little display. I can access the full addresses in the debug window with some SMSG command, but not from the GUI itself. This happens when an address line is too long, it just ends with ... 

The address above has been blurred. 

Here how it looks at the end of a line that's too long: 
screen shot 2018-12-04 at 2 57 12 pm

AllienWorks:
- https://particl.atlassian.net/browse/PD-497
- expanded the Shipping area in Order to avoid address cropping
- closes #1303 

screenshot from 2018-12-06 11-34-35

### DIFF
--- a/src/app/market/_order.scss
+++ b/src/app/market/_order.scss
@@ -98,20 +98,6 @@
       border-radius: $radius;
     }
   }
-  .detail {
-    .order-history {
-      table {
-        width: 100%;
-        border-collapse: collapse;
-      }
-      .status {
-        font-size: 11px;
-        padding: 0;
-        background: none;
-        text-align: left;
-      }
-    }
-  }
   .action-buttons {
     .single .mat-icon {
       margin: 0;
--- a/src/app/market/buy/checkout-process/checkout-process.component.html
+++ b/src/app/market/buy/checkout-process/checkout-process.component.html
@@ -293,8 +293,8 @@
         </tbody>
       </table>
 
-      <div class="order-detail" fxLayout fxLayoutAlign="space-between start" fxLayoutGap="20px">
-        <div class="shipping-overview" fxFlex="0 0 250px">
+      <div class="order-detail" fxLayout fxLayoutAlign="space-between start" fxLayoutGap="35px">
+        <div class="shipping-overview" fxFlex="1 1 100%">
           <div class="title">Shipping address</div>
           <app-shipping-address [shippingAddress]="shippingFormGroup.value"></app-shipping-address>
         </div>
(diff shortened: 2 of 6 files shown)


zaSmilingIdiot (changes requested):
This definitely widens the space to display the address. However, when the user is using a non-hi-def resolution, and/or the address is very long, it still cuts the address off. Consider for example, an address containing an office name, within a named building, in a named complex in an office park, on a long-ish named road.  - I can provide a real-life example of this if necessary :)

Playing around with the display of the address using the changes in this branch, I added in the following code changes... the idea being to break the address information into multiple lines, as well as wrap the address information on a single line if it so happens to exceed the available space:

Modify shipping-address.component.html to become:
```
<p class="address">
  <span>{{ shippingAddress?.firstName || '' }} {{ shippingAddress?.lastName || '' }} </span> <br>
  <span *ngIf="shippingAddress?.addressLine1">{{ shippingAddress?.addressLine1 || '' }} <br> </span>
  <span *ngIf="shippingAddress?.addressLine2">{{ shippingAddress.addressLine2 }} <br> </span>
  <span *ngIf="shippingAddress?.city">{{shippingAddress?.city || '' }} <br> </span>
  <span>{{ state }}</span> <br>
  <span *ngIf="shippingAddress?.zipCode">{{ shippingAddress?.zipCode || '' }} <br> </span>
  <span>{{ country }}</span>
</p>
```

Add the following into shipping-address.component.scss:
```
.address {
  width: 100%;
  word-break: break-word;
  white-space: normal;
}
```

@AllienWorks The above is more to demo what I was thinking - not sure its the best way to do this... might need your handiwork to touch up and/or make right :) But I do think we need to ensure that the text always wraps its container instead of potentially being "cut off" for some resolutions.

zaSmilingIdiot:
References issues: 
github: https://github.com/particl/particl-desktop/issues/1303
jira: https://particl.atlassian.net/browse/PD-497

AllienWorks:
@zaSmilingIdiot good idea, I've used some of it to re-layout the shipping section a bit.

Went with table in the end & labeled entries, this way it should be absolutely clear for both parties what is what (and what belongs where).

Fixed the same issue in Buy > Summary (in Checkout flow).

screenshot from 2018-12-07 12-13-15
screenshot from 2018-12-07 12-13-42

Since we've switched to this layout, we might not need these: https://github.com/particl/particl-desktop/pull/1304/commits/8ac7b9c8f65afb002c1518f5a5ea3d5ea4f0f612#diff-c1d190a7c0b1e1ed9fa0bd0c69bbfd37R23 - lines 23 & 30 (if some good soul can check that)

zaSmilingIdiot (approved):

Issue #1285, opened 2018-11-29
Listing details (description, escrow etc) absent in buy tab


kasprrr:
After i add my wanted item to the cart and go to the buy tab, there is no way to retrieve the description and escrow details of the item. I would expect to be able to click on the item in the buy tab for the details, however this is not the case.

I have to navigate back to Listings tab in order to retrieve this information and i should't have to.

HenkSwardt:
Good day @kasprrr 

Thank you for logging this. 

With regard to displaying Escrow and Shipping costs, we do cater for this in the cart. If you go to the main listings page, there is an option at the top right of the screen that allow you to select your country. By having this selected the listings page prices will update to include the shipping price in the total, depending on Local to your selected county or international. 

This then also adds more detailed pricing to your cart/checkout process, where you now see the shipping total in the main cart view and a calculation/breakdown of item price, shipping price and escrow price in the last overview page before continuing with placing the order.   

I do however like the idea of having another look at the description, @zaSmilingIdiot, can we maybe see if it will be possible to display the description as a tooltip on hover?

zaSmilingIdiot:
I think the general idea here is that in a number of e-commerce platforms, the basket contains a summary of the each item (not the full description, but then also a link back to the item listing in the store. That way a buyer can view the product details again if they want to, without the manual effort of finding the specific listing item again.

Instead of just the description on hover, I think a better idea might be to have a link next to each item in the basket (or even the item title), which when clicked on would display the entire listing item modal again, similar to having clicked the listing item on the main cart page). Your thoughts on that?

HenkSwardt:
Thank you for the feedback, 

I think that will also fit in well.

kasprrr:
"I think the general idea here is that in a number of e-commerce platforms, the basket contains a summary of the each item (not the full description, but then also a link back to the item listing in the store."

This exactly 👍

zaSmilingIdiot:
Has been implemented a described above.

PR #1320, opened 2018-12-13
added listing details to listings on buy page

EagleEye25:


zaSmilingIdiot:
@EagleEye25 please resolve the merge conflicts :)

src/app/market/buy/checkout-process/checkout-process.component.ts
```diff
@@ -374,6 +375,17 @@ export class CheckoutProcessComponent implements OnInit, OnDestroy {
     this.stepper.linear = false;
   }
 
+  openListing(listing: any) {
+    if (new Date().getTime() < listing.listing.expiredAt) {
```
zaSmilingIdiot: Rather than duplicating the expiry check here, use the `this.checkExpired()` function (which does this check already)

src/app/market/buy/checkout-process/checkout-process.component.html
```diff
@@ -48,11 +48,13 @@
           <tr *ngFor="let listing of cart?.listings"
               [ngClass]="{expired: checkExpired(listing)}">
             <td class="item" fxLayout="row" fxLayoutGap="10px" fxLayoutAling="start center">
-              <div class="image" fxFlex="0 0 75px">
-                <img src="{{listing.imageCollection?.featuredImage?.thumbnail}}">
+              <div class="image" fxFlex="0 0 115px">
+                <img src="{{listing.imageCollection?.featuredImage?.thumbnail}}" (click)="openListing(listing)"
```
zaSmilingIdiot: If the listing has been disabled due to an error message, then there is still the indication to the user to "click here for more information". Should such "help" indicators not be removed when the action indicated has been disabled/removed?

src/app/market/buy/checkout-process/checkout-process.component.html
```diff
@@ -48,11 +48,13 @@
...
+                  matTooltip="Click for more info" matTooltipPosition="above">
               </div>
               <div class="info" fxFlex="1 1 100%" fxLayout="column" fxLayoutAlign="center start">
-                <div class="title">
+                <div class="title" (click)="openListing(listing)"
+                  matTooltip="Click for more info" matTooltipPosition="above">
```
zaSmilingIdiot: If the listing has been disabled due to an error message, then there is still the indication to the user to "click here for more information". Should such "help" indicators not be removed when the action indicated has been disabled/removed?

zaSmilingIdiot (changes requested):

EagleEye25:
@AllienWorks i have found an issue with the scss, this is only present on high resolution monitors.
on high res screens the lines that are showed to indicate that the item is not accessible do not show, but on normal 1080p monitors it does:
screen shot from 4k monitor 
ss 2018-12-18 at 12 44 01
if it would be possible to find a work around this, that would be great.
i would also like to ask if you would be able to change the cursor to the normal cursor when hovering over the expired item. so it doesnt look like you can click on it when its expired.
please and thank you :D

EagleEye25:
just realized that on your screen, you will still see the lines from the screenshot, this is how it looks from my end (picture taken from my phone):

AllienWorks:
Will look into it right away :)

AllienWorks:
@EagleEye25 I've fixed the layout issue with images (probably a bad merge) & styled the Expired ones a bit better. However as I don't have any expired listings in my cart, I can't really check.

Also I can't reproduce the issue with stripes you're describing. I've tried smaller and larger viewport widths, seem to work all right.

zaSmilingIdiot:
I think the problem with the stripes is 4k resolution related (rather than
dimensional changes). Specifically, I think that the stripes are thin
enough that the pixels used to respresent them on a 4k resolution are
"lost". Zooming will probably show the lines correctly, but a 1 pixel wide
line, for example, is so thin on a 4k resolution that it is not visible.
Hope that makes some sense (my terminology may be somewhat off) :)

EagleEye25:
I'll check it out tomorrow when I get to the office, to see if the issue still persists

vikas-cis-zz:
@zaSmilingIdiot accept that comment I think the pr looks good to me :)

AllienWorks:
> @zaSmilingIdiot: I think that the stripes are thin enough that the pixels used to respresent them on a 4k resolution are "lost". Zooming will probably show the lines correctly, but a 1 pixel wide line, for example, is so thin on a 4k resolution that it is not visible.

I don't think this is possible. 1px is always 1px, no matter how large your resolution is. Sure, it will be smaller, but I doubt 4k monitors lose details like that. Moreover, the stripe is ~2px wide.

I suppose you're still seeing the borders around these comments on Github, right? Those are 1px wide..

AllienWorks:
@EagleEye25 Another thing that comes to my mind is your monitor brightness/contrast. You Particl Desktop looks kinda "overburnt".. do you maybe have some non-default profile set up? You know, those "Gaming/Movie" presets etc.

Can you post your screenshot directly (not a photo)? It might answer this..

zaSmilingIdiot:
The lines are definitely there, and the 1 pixel width being "lost" is more accurately described as being: its still visible but the grey fades into the surrounding white, losing the grey colouring and making it appear as a plain white background (so "losing" the thin striping).

Again, the lines are definitely present, and choosing a darker colouring for the lines, or changing the background colour of the row provides enough contrast for the grey lines to be shown.

AllienWorks:
Hold the phone here.. can you see the stripes in [this screenshot](https://user-images.githubusercontent.com/28861857/50149171-1d011800-02c3-11e9-8549-18aba7fc72a2.png)?
Answer to this will easily reveal the issue. It's not _only_ the stripes there; even the background of expired items is darker (same as generic app background, the light grey).

EagleEye25:
@AllienWorks on the 4k screen i cant see any difference, its just white, but when i take it over to the 1080p screen i can see the stripes.

zaSmilingIdiot (approved):

zaSmilingIdiot:
Implements https://github.com/particl/particl-desktop/issues/1285

PR #1336, opened 2018-12-21
fixed the fact that you couldnt leave the error screen

EagleEye25:


EagleEye25:
there will be some styling for the button that has to be done.
@AllienWorks if you wouldnt mind doing this for us :D
ss 2018-12-21 at 11 50 44
this is what it currently looks like :(

AllienWorks:
@EagleEye25 I've styled it, it's ready for review.
Btw the last screen appears only with errors, right?

EagleEye25:
@AllienWorks, this shows both when there are errors and when there isnt

EagleEye25:
@AllienWorks yes, this only shows when there is errors
@zaSmilingIdiot  this is ready for Review

zaSmilingIdiot (approved):
I'm not a big fan of setting global values for single state (step) content within a component. But it does work here, so for the sake of moving things along, it good enough :)

PR #1339, opened 2018-12-29
graph changes and added weight to calculate graph

vikas-cis-zz:
WIP, This needs to be tested again when cube done his changes from marketplace

Thanks


vikas-cis-zz:
@AllienWorks can you please do small css fix for the total coins showed on the proposal, we need it to look the same way as what the other values on overview page and from mine side I did show the values in three  span I think just a class add required . Let me know if you need more information 

Thanks 
screen shot 2019-01-29 at 6 41 32 pm

AllienWorks:
This is what I have so far (faked it a bit for the purpose of styling):

screenshot from 2019-01-30 11-32-18

But I think that it would make more sense to replace this with the "**XX % of coins voted**" as originally intended. Absolute amount is weird in this case, as supply is growing each day..

zaSmilingIdiot (approved):

PR #1344, opened 2019-01-07
implemented fix for item expired on last step

EagleEye25:


src/app/market/buy/checkout-process/checkout-process.component.ts
```diff
@@ -397,3 +402,7 @@ export class CheckoutProcessComponent implements OnInit, OnDestroy {
   }
 
 }
+
+export enum errorType {
```
zaSmilingIdiot: I don't think this needs to be exported, since its only used in this component. Also, to fix the warnings, it might be helpful to declare this near the top of the file instead of at the bottom.

src/app/market/buy/checkout-process/checkout-process.component.ts
```diff
@@ -306,8 +312,13 @@ export class CheckoutProcessComponent implements OnInit, OnDestroy {
       this.snackbarService.open('Order has been successfully placed');
       this.onOrderPlaced.emit(1);
     }, (error) => {
+    if (error === errorType.itemExpired) {
```
zaSmilingIdiot, line 315: Since we can streamline this:

```
    if (error === errorType.itemExpired) {
      this.resetStepper();
    }
    this.snackbarService.open(error, 'warn');
    this.log.d(`Error while placing an order`);

```

The above works a little better than the if { ... } else { ... } construct

zaSmilingIdiot (changes requested):
Works good enough, just one final change needed.

zaSmilingIdiot (approved):

PR #1348, opened 2019-01-09
created check on bid.order for validating items

EagleEye25:


zaSmilingIdiot:
Just commenting here to indicate the reason it has not yet been merged:
The current change returns the user to step 1 in the checkout process if a listing item expires while on step 3 (the final step). The user is then prompted to remove the expired item(s) from their cart in order to continue. If other items existed in their cart, then when navigating to step 2 after removing the expired item(s), the address inputs are now blank (although if a pre-saved address was selected, this dropdown is still selected). 

This is not an issue with this code change specifically, but it does need to be resolved when merging this change in in order to prevent the problem from occurring.
The fix for this scenario is being worked on at the moment.

PR #1350, opened 2019-01-10
Fixes creation of cloned listings

zaSmilingIdiot:
Cloned templates were referring to the original template, and were thus unable to be saved (or subsequently published) if the original template had a published listing associated with it. Which meant cloning a template would only be possible for an unpublished listing, which is somewhat pointless. 

The clone button now properly sets up a new template using the details of the cloned template, allowing for a new template and subsequent listing item to be created.

Also includes a fix for displaying the loading overlay when a new template is created (previously the loading overlay was only displayed on publishing of the template or when a saved template was being updated, but not on the saving of a new template).

References back to: https://particl.atlassian.net/browse/PBD-159

### DIFF
--- a/src/app/core/market/api/template/image/image.model.ts
+++ b/src/app/core/market/api/template/image/image.model.ts
@@ -28,6 +28,17 @@ export class Image {
     return (image && image.dataId) || './assets/images/placeholder_1-1.jpg'
   }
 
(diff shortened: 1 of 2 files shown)


zaSmilingIdiot:
PR made to incorrect branch

PR #1363, opened 2019-01-16
Alpha/warning messages

AllienWorks:
**Video preview:** https://i.imgur.com/Itlfxnq.mp4

### DONE

- fixed layout for "new release notification"
- added Alpha warning message on top of PD
- ..which triggers modal with all the details
- ..where I added the "Keep in mind" list to summarize (@particl/marketing-team is that OK?)

### TODO

- [ ] @vikas-cis condition for `main-view.component` to have CSS class `with-top-notification` only when messages are present (messages: new releases / this alpha warning - all in `release-notification.component`) - see [this comment](https://github.com/particl/particl-desktop/commit/68c4501b7634f1875aca66a8279f9576c188abe3#diff-266938d3389f8be359f9ac327f60a7eeR1)
- [x] @particl/marketing-team proofreading of the whole thing (I've changed a few words here and there)
- [ ] @AllienWorks when we have a final text, add it in TOS as well!

screenshot from 2019-01-16 15-22-46
screenshot from 2019-01-16 15-22-54


LiteBit:
reads good :) 
my only comment would be tweaking this part
`Neither the Particl Foundation nor its team of developers will be held accountable for any financial loss or any form of privacy and security breach.`

The Particl Foundation doesn't **have** a team of developers, rather it does fund some of the development on Particl Project. 

Maybe it could tweak to 
`Neither the Particl Foundation nor the developers working on the Particl Project will be held accountable for any financial loss or any form of privacy and security breach.`

cryptoguard:
Just really small tweak proposals, I think it really does the job well as-is too: 

- Here -> "Read these details first!" I would maybe not put an ! after "first" as it appears in the middle of the sentence and feels weird, but also I understand why it's there to put emphasis so I'm 50/50 on it.

- Maybe put the banner as "**Click here to read all the details first**, and use at your own risk!" instead of "Read these details first! and use it at your own risk!". That would highlight more the fact that this red text can be clicked

- "To reduce these risks, consider creating **a** new wallet with **a** small amount of PART for testing the marketplace" Added a's

- "All marketplace transactions are executed over mainnet and are real - Buyers are paying with real money and vendors should send out the goods!" 

- "The marketplace is not yet 100% private"

AllienWorks:
Updated according to your feedback, thanks guys!

screenshot from 2019-01-17 11-35-58

vikas-cis-zz:
@AllienWorks  I have added the implementation of the release notification dynamic class, please have a look and let me know 

Thanks

AllienWorks:
> @vikas-cis I have added the implementation of the release notification dynamic class, please have a look and let me know

I don't see anything :) so I guess it's working.. but it looks like you implemented only displaying the message when there's a new update available, right?

We also need to check if the Alpha warning message should be visible (the whole point of this PR). I guess we'll control that manually somewhere @zaSmilingIdiot? Or maybe some condition "if on mainnet & version includes _beta_" or something..?

zaSmilingIdiot:
Hmmm... I wasnt certain whether this was ready yet. I'll take a look over it in the a couple of hours and see if anything is outstanding and get it worked in so we can get rhis merged.

vikas-cis-zz:
> We also need to check if the Alpha warning message should be visible (the whole point of this PR). I guess we'll control that manually somewhere @zaSmilingIdiot? Or maybe some condition "if on mainnet & version includes _beta_" or something..?

Hye @AllienWorks which message you want to show ?

Thanks

AllienWorks:
> @vikas-cis Hye @AllienWorks which message you want to show ?

See the [screenshot in this PR](https://user-images.githubusercontent.com/1965795/51255040-b430bc80-19a2-11e9-9618-a955ced910b3.png) – that is the "alpha warning message", which we need to show.

It's [here in the code](https://github.com/particl/particl-desktop/commit/68c4501b7634f1875aca66a8279f9576c188abe3#diff-c692952549e73e66f292a6e04b55c18cR24).

vikas-cis-zz:
Hye @AllienWorks I have added some stuffs for checking the alpha message please check and let me know if that make sense 

Thanks

AllienWorks:
Just pushed added warnings to the buy flow; in:

- confirm bid modal
- accept bid modal
- confirm payment window
- confirm delivery

See [screenshots](https://imgur.com/a/gfhW2VA) for illustration. @particl/marketing-team can I get some proof-reading again pls? :)

AllienWorks:
Since notification for new releases are moving to version numbers in bottom left (see https://github.com/particl/particl-desktop/pull/1378 PR), **we should remove the "new release available" from `release-notification` component**. Ideally rename the component to `announcement-notification` & keep there just the Alpha mainnet warning.

src/app/core-ui/main/announce-notification/announcement-notification.component.ts
```diff
@@ -46,4 +51,14 @@ export class ReleaseNotificationComponent implements OnInit, OnDestroy {
     return (parseFloat(this.currentClientVersion) < parseFloat(this.latestClientVersion));
   }
 
+  // check for beta version
+  isAlpha(): boolean {
```
zaSmilingIdiot: There's no need for this method. Set the value of the variable 'mainnet' in onInit(), and use that in the template, rather than a function call.

src/app/core-ui/main/announce-notification/announcement-notification.component.html
```diff
@@ -0,0 +1,25 @@
+<!-- Replace with the new changes that allien made on different branch -->
```
zaSmilingIdiot: If this commented code is no longer necessary, then remove it. The card further down seems to be doing the same thing.

src/app/core-ui/main/announce-notification/announcement-notification.component.ts
```diff
@@ -46,4 +51,14 @@ export class ReleaseNotificationComponent implements OnInit, OnDestroy {
     return (parseFloat(this.currentClientVersion) < parseFloat(this.latestClientVersion));
```
zaSmilingIdiot: Its better to remove this whole method. Once a new version has been successfully downloaded, check its version against the current version and set a boolean value. Use this boolean value in the template to determine whether a newer version is available.

zaSmilingIdiot: You may also want to verify the parseFloat() of the string actually works. A false value may be generated in an example scenario where
this.currentClientVersion = '2.1.1'
and
this.latestClientVersion = '2.1.2'

In which case:
parseFloat('2.1.1') < parseFloat('2.1.2') === 2.1 < 2.1 === false
Which means that the user would not be notified of a newer update.

src/app/core-ui/main/main-view.component.ts
```diff
@@ -161,6 +162,11 @@ export class MainViewComponent implements OnInit, OnDestroy {
     return sec;
   }
 
+  // check for beta version
+  isAlpha(): boolean {
```
zaSmilingIdiot: There's no need for this method. Use a single variable which is initialized in the onInit() method, and update the template to make use of that variable.

src/app/core-ui/main/announce-notification/announcement-notification.component.ts
```diff
@@ -2,23 +2,28 @@ import { Component, OnDestroy, OnInit } from '@angular/core';
...
 @Component({
-  selector: 'app-release-notification',
-  templateUrl: './release-notification.component.html',
-  styleUrls: ['./release-notification.component.scss']
+  selector: 'app-announcement-notification',
+  templateUrl: './announcement-notification.component.html',
+  styleUrls: ['./announcement-notification.component.scss']
 })
-export class ReleaseNotificationComponent implements OnInit, OnDestroy {
+export class AnnounceMentNotificationComponent implements OnInit, OnDestroy {
```
zaSmilingIdiot: s/AnnounceMentNotificationComponent/AnnouncementNotificationComponent/g

There's a weird capital letter 'M' in 'Announcement'

src/app/core-ui/main/announce-notification/announcement-notification.model.ts
```diff
@@ -1,5 +1,5 @@
 
-export class ReleaseNotification {
+export class AnnounceMentNotification {
```
zaSmilingIdiot: s/AnnounceMentNotificationComponent/AnnouncementNotificationComponent/g

There's a weird capital letter 'M' in 'Announcement'

src/app/core-ui/main/announce-notification/announcement-notification.component.ts
zaSmilingIdiot: On second thought, is this check actually necessary for this component?

src/app/core-ui/main/announce-notification/announcement-notification.component.ts
```diff
@@ -33,7 +38,7 @@ export class ReleaseNotificationComponent implements OnInit, OnDestroy {
 
   getCurrentClientVersion() {
```
zaSmilingIdiot: It is probably not necessary for this method to exist. The component exists here to simply inform users of notifications. Thus its not necessary to be checking for new version releases in this component.

src/app/core-ui/main/announce-notification/announcement-notification.component.html
```diff
@@ -0,0 +1,25 @@
...
+</mat-card> --><!-- .notification -->
+
+
+<!-- Alpha mainnet warning -->
+<mat-card class="notification alert" *ngIf="isAlpha()"><!-- *ngIf="" -->
+  <mat-card-title class="title">
+    The Particl Marketplace alpha is still in development and not 100% private yet &ndash; use it at your own risk!
+    <p *ngIf="latestClientVersion === currentClientVersion">You are on the latest version</p>
+
+    <p *ngIf="isNewUpdateAvailable()">You are on an outdated version, please click to update</p>
```
zaSmilingIdiot: Is any of this HTML necessary, if the component exists to inform users of new messages (or currently just the alpha/beta warning which has already been included above)? This can likely be removed.

zaSmilingIdiot (changes requested):
So a couple of things:
- firstly, the PR was branched from market rather than market-beta, which is why there are all the additional files. Which is not a problem, just mentioning it :) There were fortunately no significant additional changes in the market branch at the time this was created.

- It might be better to create a single function that determines the type of release. Then use that everywhere you need it. It means that in the future if the format of the release version needs to be changed/updated, its a single location to do so. So better maintainability.

- Would the modals not benefit from a similar conditional check to show/hide the notification if the release is a alpha/beta release? This would mean that no changes are necessary when a final version is released (the warnings are no longer visible). It also means that future alpha/beta release updates would automatically have the warning visible again.

- I understand the reason behind the change of the component naming from ReleaseNotification to AnnouncementNotification. In such a case though, as I've mentioned in the comments, rather remove the unnecessary checks and validations for new client versions. Leave the component doing exactly its purpose, and only its purpose.

src/app/core-ui/main/announce-notification/announcement-notification.component.ts
vikas-cis-zz: Hye @zaSmilingIdiot  I was thinking to use the same component on mainview component that the PR1378 refers to that is the reason I doesn't removed the code tho I will fix the small stuffs that you mentioned

src/app/core-ui/main/announce-notification/announcement-notification.component.ts
vikas-cis-zz: I initially thought regarding that extra M to be small but thought it will better to be in Capital :D :D :D

AllienWorks:
> firstly, the PR was branched from market rather than market-beta, which is why there are all the additional files. Which is not a problem, just mentioning it :) There were fortunately no significant additional changes in the market branch at the time this was created.

Sorry for this. I initially did it like that, since you said `market` and `market-beta` are the same anyway. Learned my lesson :)

> It might be better to create a single function that determines the type of release. Then use that everywhere you need it. It means that in the future if the format of the release version needs to be changed/updated, its a single location to do so. So better maintainability.
> 
> Would the modals not benefit from a similar conditional check to show/hide the notification if the release is a alpha/beta release? This would mean that no changes are necessary when a final version is released (the warnings are no longer visible). It also means that future alpha/beta release updates would automatically have the warning visible again.

Fully agreed, let's do that.

> I understand the reason behind the change of the component naming from ReleaseNotification to AnnouncementNotification. In such a case though, as I've mentioned in the comments, rather remove the unnecessary checks and validations for new client versions. Leave the component doing exactly its purpose, and only its purpose.

Absolutely - leave the component doing its purpose. But since version checks are moving elsewhere, "release notification" is not exactly correct IMO. As I see it, this message space can now be used for whatever messages we need to show (not just release-related), so I'd change the naming of the component.

zaSmilingIdiot:
> Absolutely - leave the component doing its purpose. But since version checks are moving elsewhere, "release notification" is not exactly correct IMO. As I see it, this message space can now be used for whatever messages we need to show (not just release-related), so I'd change the naming of the component.

Yep, I'm agreeing with you :) The old code that did the check for the release updates was still included in the component though - I was just asking that if the component is becoming a NotificationComponent then that release checking code should probably be removed from it.

In saying that though, I still think the current NotificationComponent is pretty rigid and fixed at being only a alpha/beta warning message. The actual notifications should probably be obtained from somewhere specific, and the component should probably be a bit more generic to general notifications. So for example, define a generic data structure with a title, subtitle, and action. The card that is the notification then only displays the parts of the message that have been defined; eg: if no subtitle has been provided, then don't display a subtitle.
The messages are then obtained from a service. Which would allow for future messages to be generated easily enough, without rewriting the component - just push the message into the queue, and it gets delivered to the notification component.

zaSmilingIdiot (dismissed):

zaSmilingIdiot:
Hmmm... still needs one or 2 quick fixes...

zaSmilingIdiot (approved):

PR #1364, opened 2019-01-16
Fix upload images

EagleEye25:


zaSmilingIdiot:
Just commenting here that this fixes the problem where attempting to include the same image consecutively fails via the open file dialog, but including the same image after the file dialog has opened at least once (for either another image or no image in the case that the dialog is cancelled) works.

zaSmilingIdiot (approved):

PR #1367, opened 2019-01-17
- Proposal service test-case.

cis-shubham-t:


zaSmilingIdiot (approved):
This was from accidently reverting these changes out.

PR #1378, opened 2019-01-25
Version check indicators

AllienWorks:
Since we won't use the top message panels for notifying users about new versions, this PR moves this to bottom left, directly to version numbers (makes more sense anyway!):

version-check

### Done

- styled 3 possible cases:
  - (green) current version, all OK
  - (blinking red) version in use is outdated, user asked to update it
  - (orange) unable to check which version is latest (network issues etc.)

### Todo

- [ ] implement the checks and switch classes & tooltips respectively
- [ ] remove this version check functionality from top message panel (where we currently have the Alpha mainnet warning) @ `release-notification` component (and possibly rename the component to something else, like `announcement-notification`)


HenkSwardt:
Looking good @AllienWorks 

Thanks

vikas-cis-zz:
Hye @AllienWorks the removal of the component and renaming it has been done in other #1363 

Thanks

src/app/market/listings/listings.component.ts
```diff
@@ -185,9 +185,11 @@ export class ListingsComponent implements OnInit, OnDestroy {
   }
 
   loadNextPage() {
-    let nextPage = this.getLastPageCurrentlyLoaded(); nextPage++;
```
zaSmilingIdiot, line 188: @vikas-cis What does this change relate to? ie: What is this change fixing? It doesn't seem to be related to any of the other changes.

vikas-cis-zz: @zaSmilingIdiot yeah it was just kind of weird thing that happen when open your dev tool and when there are no listing then you have scroller on page and at some moment you scroll the page and you see the weird syntax error which has been fixed via the above fix.

vikas-cis-zz:
@zaSmilingIdiot  there is a reason why I didn't remove the component coz it is already done in #1363  :D :D

zaSmilingIdiot:
> @zaSmilingIdiot there is a reason why I didn't remove the component coz it is already done in #1363 :D :D

Sure, I understand that... however, if I merge this PR and for some reason there are issues with the other PR that prevent it from being merged immediately, then we effectively have 2 components doing the same thing. So each PR is evaluated separately, thus evaluation of this PR is done separately from the other one.

Consider that there may also be multiple reviewers - if one person reviews this PR and another reviews the other PR, then you're forcing them to communicate using some side channel or somehow define a single PR or external system to discuss merging strategy. That introduces unnecessary complexity, so its one PR at a time :)

vikas-cis-zz:
@zaSmilingIdiot  Ahh!! I see no problem 

Thanks

zaSmilingIdiot (approved):

PR #1383, opened 2019-01-28
Ensures addressindex flag is included on daemon startup

zaSmilingIdiot:
Checks if the addressindex flag is added as a cli flag. If not, read the particl.conf: if its not set or set to disabled, then write change to the particl.conf file and ensure that -reindex-chainstate flag is used when starting the daemon up.

### DIFF
--- a/modules/daemon/daemonConfig.js
+++ b/modules/daemon/daemonConfig.js
@@ -8,12 +8,29 @@ const _options    = require('../options');
 const conFilePath = path.join( cookie.getParticlPath(_options.get()), 'particl.conf');
 const SAFE_KEYS = ['addressindex'];
 
+const isArray = function(obj) {
+  return Object.prototype.toString.call(obj) === '[object Array]';
+}
+
+const isObject = function(obj) {
+  return Object.prototype.toString.call(obj) === '[object Object]';
+}
+
+deepClone = function(obj) {
+  let retVal;
+  try {
+    retVal = JSON.parse(JSON.stringify(obj))
+  } catch(err) {
+    retVal = undefined;
+  };
+  return retVal;
+}
 
 const formatSettingsOutput = function(rawConfig) {
   let formattedConfig = {};
 
     let sections = [];
-    if (Object.prototype.toString.call(rawConfig) === '[object Object]' && Object.prototype.toString.call(rawConfig.items) === '[object Array]') {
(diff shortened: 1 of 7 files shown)


modules/daemon/daemonConfig.js
```diff
@@ -0,0 +1,198 @@
...
+
+const conFilePath = path.join( cookie.getParticlPath(_options.get()), 'particl.conf');
+const SAFE_KEYS = ['addressindex'];
+
+
+const formatSettingsOutput = function(rawConfig) {
+  let formattedConfig = {};
+
+    let sections = [];
+    if (Object.prototype.toString.call(rawConfig) === '[object Object]' && Object.prototype.toString.call(rawConfig.items) === '[object Array]') {
```
kewde: Build a few helper functions isObject() and isArray()?

modules/daemon/daemonConfig.js
```diff
@@ -0,0 +1,198 @@
...
+  let isGlobalConfig = true;
+
+  for (let ii = 0; ii < items.length; ii++) {
+    const section = items[ii];
+    const hasNodes = Object.prototype.toString.call(section.nodes) === '[object Array]';
+
+    if ('name' in section) {
+      // section delimiting (could be global config item or blank line as well)
+      let name = section.name;
+      if (section.name.length && hasNodes && section.nodes.length) {
```
kewde: move `&& section.nodes.length` into hasNodes?

kewde (approved):
Seems like a good base for the settings PR.

modules/daemon/daemonConfig.js
```diff
@@ -0,0 +1,198 @@
...
+  return formattedConfig;
+}
+
+const readConfigFile = function () {
+  log.debug('Attempting to read particld config from: ', conFilePath);
+  if (fs.existsSync(conFilePath)) {
+    try {
+      const p = new iniParser();
+      const result = p.parse(fs.readFileSync(conFilePath, 'utf-8'));
+      return JSON.parse(JSON.stringify(result));
```
kewde: Maybe write helper function clone()?

zaSmilingIdiot:
> Seems like a good base for the settings PR.

Yep, this was actually the basis for my changes in the Settings PR. I ripped it from there, fixed up a little bug that I discovered, and here she is :) (which is also why all the TODOs still exist).

PR #1389, opened 2019-01-30
fixed sell tab issues

EagleEye25:


EagleEye25:
ticket https://particl.atlassian.net/browse/PBD-197

src/app/core/market/api/template/template.service.ts
```diff
@@ -55,7 +55,13 @@ export class TemplateService {
 
   search(page: number, pageLimit: number, sort: string, profileId: number, category: string,
     searchString: string, hashItems: any): Observable<Array<Template>> {
-    const params = ['search', page, pageLimit, 'DESC', sort,  profileId, searchString, category, hashItems];
+      let direction;
+      if (sort === 'TITLE') {
```
zaSmilingIdiot, line 60: This is not the correct place to be doing this, although the correct action would be to have the necessary radio buttons that allow for user selection of the sort direction. While this is still something that needs to be implemented correctly, please leave a comment here with a TODO action to ensure this is re-worked and not forgotten about in some issue ticket comment.

vikas-cis-zz: @zaSmilingIdiot I was just wondering why it has been changed as it was previously managed on component side, as we got listings when sorting via title in descending order.
Thanks

zaSmilingIdiot: @vikas-cis We don't yet provide an option for the user to select which sort direction (ASC or DESC) they want the results to occur in. However, the marketplace api does allow for either to be used, and the latest updates to the marketplace at least have the sorting direction working correctly.

So why not just leave the sort order to be DESC always for now? Mostly because the default ordering by date should probably remain descending (so that it remains newest -> oldest), but the ordering direction for titles is more natural to default to ascending (for example A -> Z).

zaSmilingIdiot (approved):

src/app/core/market/api/template/template.service.ts
vikas-cis-zz: @zaSmilingIdiot  yeah that is why I reversed the listing in the component when it is title coz we need to show the listing in desc(newest to older) order I believe .

Thanks

PR #1391, opened 2019-01-31
fix back to sell on save template

EagleEye25:


zaSmilingIdiot:
@EagleEye25 can you provide some input as to why this would be needed?

Perhaps the wording of the 'Save' button is incorrect. The seller, when creating a template, may want to save their current changes, but still wants to provide updated values. Moreso applicable since we are providing checks that the template size fits into the resulting smsgmessage. In such a case a seller may want to save the template with a base set of text and images, check that everything is valid and working, and then publish.

But taking the user back to the listings page each time they save their changes, it becomes an annoyance to try save or update. Hence, a valid scenario is required for this (and its likely requires additional work as to implement).

zaSmilingIdiot (approved):

PR #1394, opened 2019-02-01
Fix upload image

EagleEye25:


vikas-cis-zz (commented):
we can make it more shorter 

```
validImageTypes = [  'image/jpeg',  'image/png' ];

if (validImageTypes.includes(file.type)) {
failedImgs = file.size > MAX_IMAGE_SIZE;
} if (!failedImgs) {
 do the rest I believe ?
}
```

src/app/market/sell/add-item/add-item.component.ts
```diff
@@ -132,15 +130,18 @@ export class AddItemComponent implements OnInit, OnDestroy {
     const MAX_IMAGE_SIZE = 1024 * 1024 * 2; // (2MB)
     let failedImgs = false;
     sourceFiles.forEach((file: File) => {
-      if (file.size > MAX_IMAGE_SIZE) {
-        failedImgs = true;
+      if (this.validImageTypes.indexOf(file.type) > -1) {
```
zaSmilingIdiot, line 133: Note that the MIME type of a file obtained using the File object's _type_ property makes use of the file extension to determine the mime type of the file, rather than using the first few bytes of the file to determine it accurately. 
See https://developer.mozilla.org/en-US/docs/Web/API/File/type

What this means is that anybody could rename the extension of the file that is being uploaded in order to work around the file type restriction.

zaSmilingIdiot: For reference, here is an article that indicates how to check a binary file type correctly:
https://medium.com/the-everyday-developer/detect-file-mime-type-using-magic-numbers-and-javascript-16bc513d4e1e

Other references do also exist :)

zaSmilingIdiot: Also, combine this condition into the nested if condition, such that it reads something like: 
`failedImgs = file.size > MAX_IMAGE_SIZE && isValidMimeType()` 

then change the error message of '1 or more images failed: max image size is 2MB' to something more generic for example like 'Failed to include an invalid image'

ie: no need for nested conditionals here.

zaSmilingIdiot (changes requested):

zaSmilingIdiot:
Looking back over this PR, the logic of the file uploads has also been changed such that an image failing the validation stops loading any other images. This was not the original intention. I also seem to have issues running locally this specific branch (updated to the latest particl/market-beta branch). 

I've thus implemented the required change here: 
https://github.com/particl/particl-desktop/pull/1435

Please note: I'm not in the habit of replacing PRs, but the author(s) in this case are currently involved with other activities, and I'd like to get this issue moving along, hence this action is occurring.

PR #1395, opened 2019-02-01
fix errors on resize of chart

EagleEye25:


zaSmilingIdiot:
travis seems to be on a little break... job is queued for some time now, with nothing else running. I've run the tests locally on this branch multiple times without issues. So I'm continuing with the review process now, and going to try to forget about this waste of time.

zaSmilingIdiot (approved):

PR #1402, opened 2019-02-04
removed add to cart button when item is in cart / bidding on

EagleEye25:


vikas-cis-zz:
Above PR can be close as the fix #1447 is the PR @zaSmilingIdiot 

Thanks

zaSmilingIdiot:
Seems to have been re-implemented with feqer changes in https://github.com/particl/particl-desktop/pull/1447

PR #1423, opened 2019-02-07
Listing price decimal issue

cis-shubham-t:
Task:  https://particl.atlassian.net/projects/PBD/issues/PBD-217


zaSmilingIdiot:
This seems to have been based off the wrong branch - looks like the original branch was 'market' and not 'market-beta'

Issue #1414, opened 2019-02-06
Items listed at less than one Partoshi are converted to 0 cost


CryptoBeans:
I listed an item for .000000001 PART (one tenth of the lowest possible decimal) and it appears to have published as costing 0.

I think the UI should limit the cost field to 8 decimal points and block you from posting if you try to put in a number with further decimals.

zaSmilingIdiot:
Thanks for finding this. You are indeed correct, and numeric currency input fields should be restricted to the  maximum number of supported decimal places.

HenkSwardt:
Issue has been logged for investigation.

https://particl.atlassian.net/browse/PBD-217

zaSmilingIdiot:
Has been fixed, and is included in the next development testnet release.

Issue #1415, opened 2019-02-06
One Partoshi cost item is not displaying correctly


CryptoBeans:
I listed an item for .00000001 PART. The first image is from the listings page: I can see the cost is 1e-8 which is technically correct but not pleasant to see. However, upon clicking the item (second image) it displays a cost of 0 rather than the correct amount. 

![image](https://user-images.githubusercontent.com/38863166/52360254-34c26600-2a01-11e9-8440-31248315454e.png)

![image](https://user-images.githubusercontent.com/38863166/52360293-43a91880-2a01-11e9-87c4-824af79eae5c.png)

zaSmilingIdiot:
Thanks for finding this.

HenkSwardt:
This has been logged for investigation. https://particl.atlassian.net/browse/PBD-217

zaSmilingIdiot:
Has been fixed, and is included in the next development testnet release.

PR #1424, opened 2019-02-07
- Template price decimal issue

cis-shubham-t:
Task: https://particl.atlassian.net/projects/PBD/issues/PBD-217

### DIFF
--- a/src/app/market/sell/add-item/add-item.component.ts
+++ b/src/app/market/sell/add-item/add-item.component.ts
@@ -26,7 +26,7 @@ import { MatDialog } from '@angular/material';
 
 
 class CurrencyMinValidator {
-  static validValue(fc: FormControl){
+  static validValue(fc: FormControl) {
     const amount: number = +fc.value;
     if ( amount >= 0 ) {
       if (amount > +amount.toFixed(8) ) {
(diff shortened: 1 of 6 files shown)


src/app/market/sell/add-item/add-item.component.ts
```diff
@@ -390,6 +390,16 @@ export class AddItemComponent implements OnInit, OnDestroy {
     if (this.keys.includes(event.key) || pasted.split('').find((c) =>  this.keys.includes(c))) {
       return false;
     }
+
+  }
+
+  decimalValidator(event: any) {
```
zaSmilingIdiot: Would this not be simpler and shorter implemented with something like
`parseFloat(event.target.value).toFixed(8)`
?

zaSmilingIdiot (changes requested):
This deals with the one place in the system that deals with currency inputs. Can you please confirm that the other location (wallet > send) has a similar limitation please?

zaSmilingIdiot:
Conclusion of this task should resolve https://github.com/particl/particl-desktop/issues/1414

zaSmilingIdiot:
This should also include resolution of https://github.com/particl/particl-desktop/issues/1415

zaSmilingIdiot (commented):
I'm leaving this open for now for 2 reasons:
1. Primary reason: Saving a currency input value of a small amount ( 1e-07 or less) on a template, and then editing the template again at a later stage displays the exponential form of the value rather than the actual value.

2. Minor reason: not really significant, but I'd like the noExponentialAmount function name to change to something a little more applicable, such as 'getAmountAsString' as an offhand example.

zaSmilingIdiot (approved):

Issue #1411, opened 2019-02-06
View > Reload causes app to go to a whitescreen


CryptoBeans:
Win64 build on Windows 10 machine. I right-clicked on the Particl icon and saw the options "View, Window, Help". I hovered over View and then clicked Reload. When I did that my main Particl Desktop screen went all white and did not seem to actually reload.

Restarting the app put everything back to normal.

zaSmilingIdiot:
Thanks for reporting this. Will try to replicate and determine the cause of the issue.

HenkSwardt:
This has been logged for investigation. https://particl.atlassian.net/browse/PBD-219

zaSmilingIdiot:
For now, we've removed the 'Reload' option from the menu.

PR #1426, opened 2019-02-07
hide the reload option from the build

vikas-cis-zz:
https://particl.atlassian.net/browse/PBD-219


vikas-cis-zz:
issue #1411 

As per the task we have hide the reload option 

thanks

zaSmilingIdiot (approved):

Issue #1418, opened 2019-02-06
Particl not working correctly through the .deb installer


SativaL:
This is the issue template. Everything is optional, but the more you supply, the better.
Please use short decriptive titles for issues.

# Describe bug/issue
when you install particl via the .deb installer and then search and run it, does not work because there is a error message coming up saying particl not found in folder /opt/particl. Something is wrong there.. 
..

# How to reproduce

Describe the steps needed to recreate the bug/issue:

1. ..
2. ..

# More information

- Operating system and version: ubuntu 18.04
- Client version: testnet build latest

zaSmilingIdiot:
We've had an internal report of the deb installer not working correctly as well, but for different reasons. Testing of this needs to be finalized. In the interim, you could perhaps use the Linux .zip file if you would like to continue with the testing.

HenkSwardt:
This has been logged for investigation. https://particl.atlassian.net/browse/PBD-218

zaSmilingIdiot:
This has been fixed - the next build will have the deb installer using the correct paths when launched via the GUI.

PR #1434, opened 2019-02-08
Fixes GUI launcher path for debian installs

zaSmilingIdiot:


zaSmilingIdiot:
This resolves https://github.com/particl/particl-desktop/issues/1418

PR #1442, opened 2019-02-11
particld boot: reindex if we shut it down improperly..

kewde:


modules/daemon/daemon.js
```diff
@@ -112,6 +113,13 @@ exports.start = function (wallets) {
       // TODO change for logging
       child.stdout.on('data', data => daemonData(data, console.log));
       child.stderr.on('data', data => {
+        const err = data.toString('utf8');
+        if (err.includes("-reindex") && attemptsToStart < maxAttempts) {
+          log.error('Corrupted block database detected, '
```
zaSmilingIdiot: I'm not sure that the "Corrupted block database detected" is necessary? We have at least one other error where a reindex is required, and its not related to corrupt database blocks. :) Might be confusing if trying to troubleshoot issues :D

kewde: Fixed

Issue #1437, opened 2019-02-08
Un-favoriting item from Favorites page doesn't remove it initially


CryptoBeans:
Extremely minor enhancement:

If I go to my Favorites page and click the Heart icon on a listing, it gives me the Toast message that the item is Removed from my favorites. However, the item stays on the Favorites page until I actually go to a different page and come back. Considering how things work on other pages (where things always update upon change) I think it would be consistent to remove that item from the list immediately upon de-favoriting.

HenkSwardt:
Thank you for logging this, we have logged an issue to investigate and resolve:

https://particl.atlassian.net/browse/PBD-225

zaSmilingIdiot:
This has been resolved, and the fix will be included in the next development testnet release.

PR #1444, opened 2019-02-11
initilizing the favorites when there is no favorite available

vikas-cis-zz:
https://particl.atlassian.net/browse/PBD-225

### DIFF
--- a/src/app/market/buy/buy.component.ts
+++ b/src/app/market/buy/buy.component.ts
@@ -47,6 +47,12 @@ export class BuyComponent implements OnInit {
   load() {
     this.favoritesService.cache.getFavorites().subscribe(favorites => {
       const temp: Listing[] = [];
+      // intialize when there is no favorites and no need for map
+      if (favorites.length === 0 ) {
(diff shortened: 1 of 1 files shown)


vikas-cis-zz:
#1437 
Thanks

zaSmilingIdiot:
This doesn't, at this point in time, actually fix the linked issue. I'm assuming that the change included here is a part of the fix, but its not a complete solution for the reported issue.

This change appears to modify the behaviour of the component when the favourite items reach nothing. But the reported issue is that when a favourite item is marked as not being a favourite (the favourite indication is unselected), then the item remain visible in th eFavourites page until the page is refreshed. In other words, the Favourites page is not automatically refreshing when a favourited item is removed.

vikas-cis-zz:
@zaSmilingIdiot I see may be I didn't read the ticket very well my bad :P :D 

Thanks

zaSmilingIdiot:
Previous commentary from me can be ignored

zaSmilingIdiot (approved):

PR #1458, opened 2019-02-18
Removes repeat calls to category list rpc.

zaSmilingIdiot:
Replaces the unnecessary calls with an in-Service BehaviourSubject.

### DIFF
--- a/src/app/core/market/market-state/market-state.service.ts
+++ b/src/app/core/market/market-state/market-state.service.ts
@@ -20,8 +20,8 @@ export class MarketStateService extends StateService implements OnDestroy {
      * Or Once improve-router branch got merged.
      */
 
-    this.register('profile', 0, ['list']);
-    this.register('bid', 0, ['search', 0, 99999, 'ASC', '*', '*', ''])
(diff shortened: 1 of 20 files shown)


src/app/core/market/market-state/market-state.service.ts
```diff
@@ -20,9 +20,8 @@ export class MarketStateService extends StateService implements OnDestroy {
      * Or Once improve-router branch got merged.
      */
 
-    this.register('category', 3 * 1000, ['list']);
-    this.register('profile', 60 * 1000, ['list']);
-    this.register('bid', 15 * 1000, ['search', 0, 99999, 'ASC', '*', '*', ''])
+    this.register('profile', 0, ['list']);
```
zaSmilingIdiot: @kewde or @vikas-cis or @cis-shubham-t 
Is there any reason we are making repeat profile and bid search repeated calls here? Should these not be included in a service and accessed when applicable? I've changed the timeout period to 0 (so that a single call is made), and do not currently observe any side effects. But I'd like to understand the thinking behind their inclusion here.

PR #1460, opened 2019-02-19
Updates listing item expiry option estimates for errors

zaSmilingIdiot:
Displays an error (instead of a 0 value) for listing item expiry option estimates where the expiry option would result in an error.


zaSmilingIdiot:
Resolves task issue at: https://particl.atlassian.net/browse/PBD-232

vikas-cis-zz:
Looks Good to me (y) (y) @zaSmilingIdiot

PR #1469, opened 2019-02-22
- Added the expiry time and processing related changes.

cis-shubham-t:
Task: https://particl.atlassian.net/projects/PBD/issues/PBD-240

### DIFF
--- a/src/app/wallet/proposals/models/proposal.model.ts
+++ b/src/app/wallet/proposals/models/proposal.model.ts
@@ -11,6 +11,8 @@ export class Proposal {
   public id: number;
   createdAt: number;
   expiredAt: number;
+  MAX_SAFE_INTEGER: number = 9007199254740991;
+  isExpiredAtValid: boolean;
 
   constructor(object: any) {
     this.title = object.title;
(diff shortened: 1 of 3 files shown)


src/app/wallet/proposals/models/proposal.model.ts
```diff
@@ -22,6 +24,13 @@ export class Proposal {
     this.id = object.id;
     this.createdAt = object.createdAt;
     this.expiredAt = object.expiredAt;
+    /**
+     * object.expiredAt is equals MAX_SAFE_INTEGER = 9007199254740991. if the proposal is not publish.
+     * (once smsg message sent and recieveed) expiry time will be change as it is actual instead of
+     * MAX_SAFE_INTEGER = 9007199254740991.
+     */
+
+    this.isExpiredAtValid = object.expiredAt < this.MAX_SAFE_INTEGER;
```
zaSmilingIdiot: This can be more safely written as:
`this.isExpiredAtValid = Number.isSafeInteger(object.expiredAt)`

Which also means that the MAX_SAFE_INTEGER const created in this file is also no longer needed. 
For future reference, the max safe integer value can be obtained via `Number.MAX_SAFE_INTEGER`

zaSmilingIdiot (changes requested):
Also just requested clarification from the author ont he reason for some of the changes included here.

zaSmilingIdiot (approved):

PR #1474, opened 2019-02-25
wrapping up the text and added hint on proposal addying

vikas-cis-zz:
https://particl.atlassian.net/browse/PBD-239

### DIFF
--- a/src/app/wallet/proposals/proposal-details/proposal-details.component.ts
+++ b/src/app/wallet/proposals/proposal-details/proposal-details.component.ts
@@ -43,10 +43,10 @@ export class ProposalDetailsComponent implements OnInit, OnDestroy {
       showLabels: false,
       donut: true,
       legend: {
-        maxKeyLength: 100,
+        maxKeyLength: 35,
         margin: {
           left: 200,
-          right: 100,
+          right: 112,
           top: 5,
           bottom: 5
         },
--- a/src/app/wallet/proposals/proposal-details/proposal-details.component.ts
+++ b/src/app/wallet/proposals/proposal-details/proposal-details.component.ts
@@ -45,41 +45,18 @@ export class ProposalDetailsComponent implements OnInit, OnDestroy {
       legend: {
         maxKeyLength: 100,
         margin: {
-          left: 0,
-          right: 0,
+          left: 200,
+          right: 100,
           top: 5,
           bottom: 5
         },
-        padding: {
-          top: 0,
-          bottom: 0,
-          right: 5,
-          left: 5
-        }
+        rightAlign: true
       },
       color: ['#02E8B0', '#ec4b50', '#108cda', '#f1cc00', '#7e6c95'], // green, red, blue, yellow, violet
       tooltip: {
         enabled: true,
         hideDelay: 0,
         useInteractiveGuideline: false
-      },
-      callback: function() {
-        d3.selectAll('.nv-series text').each(function() {
-          const text = d3.select(this);
-          let words = text.text();
-          const dy = 0;
-          text.text(null).append('tspan').text(words.substring(0, 30))
-          if (words.length > 30) {
-            words = words.substring(30, words.length);
-            d3.select(this.parentNode).append('text').text(null)
-              .attr('text-anchor', 'start')
-              .attr('class', 'nv-legend-text')
-              .attr('dy', (dy + 1) + 'em')
-              .attr('dx', '8')
-              .append('tspan')
-              .text(words)
-          }
-        });
       }
     }
   };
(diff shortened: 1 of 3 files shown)


src/app/wallet/proposals/proposal-details/proposal-details.component.ts
```diff
@@ -43,25 +43,43 @@ export class ProposalDetailsComponent implements OnInit, OnDestroy {
...
+          left: 5
+        }
       },
       color: ['#02E8B0', '#ec4b50', '#108cda', '#f1cc00', '#7e6c95'], // green, red, blue, yellow, violet
       tooltip: {
         enabled: true,
         hideDelay: 0,
         useInteractiveGuideline: false
+      },
+      callback: function() {
```
zaSmilingIdiot: This function doesn't seem to work 100% for wrapping the text. Particularly when resizing the application window (I'm noticing a significant amount of text overlap).

If this is not particularly possible or easy to implement, then lets leave it for now and address it as part of a different issue.

src/app/wallet/proposals/proposal-details/proposal-details.component.ts
```diff
@@ -43,25 +43,43 @@ export class ProposalDetailsComponent implements OnInit, OnDestroy {
...
           top: 5,
           bottom: 5
         },
-        // padding: {
-        //   top: 0,
-        //   bottom: 0,
-        //   right: 5,
-        //   left: 5
-        // }
+        padding: {
```
zaSmilingIdiot: Please do not change this, ie: leave it commented out or better yet, remove the padding object here altogether. Using an object for padding results in errors in the console. The correct way to set padding appears to be via a number value (rather than an object value).

zaSmilingIdiot (changes requested):
We may need to look at an alternative graphing library... perhaps something like the following as a possible option:
https://swimlane.gitbook.io/ngx-charts/

The text wrapping issue can be addressed via an alternative issue, if its implementation is non-trivial. The priority concern with this fix was (is) to ensure that the legend text starts from within the visible space.

src/app/wallet/proposals/proposal-details/proposal-details.component.ts
vikas-cis-zz: Hye @zaSmilingIdiot I am not which error you are getting tho the padding helps in letting the text to start after the circle point. If remove that then the  initial text will be hide and looks messy .

Thanks

src/app/wallet/proposals/proposal-details/proposal-details.component.ts
vikas-cis-zz: @zaSmilingIdiot Is that is the known issue and even if you click on dots the text field again rendered due to which it showing the breaking text in one line again which is weird for me 

Thanks

src/app/wallet/proposals/proposal-details/proposal-details.component.ts
zaSmilingIdiot: You can generate the error by ensuring that the JS console is open, and load the Proposals page. An error is generated for each graph. Additional errors can be generated by resizing the window.

Have a look over here: http://krispo.github.io/angular-nvd3/#/donutChart
Change the radio button from 'Basic' to 'Extended' and then expand: cahrt -> legend. The 'padding' value accepts a numerical value not an object. If the padding is required, then try to replace it with a number rather than an object. I suspect that if yo uneed further fine grain control of the padding values that you may need to use the chart -> styles -> object, or an alternative solution (css styles for the svg elements? ).

src/app/wallet/proposals/proposal-details/proposal-details.component.ts
zaSmilingIdiot: Okay, so because its a known issue, then we either need to correct it or remove it. 

I don't see a need to waste unnecessary amounts of time to try to resolve this right now. If its not resolvable in a few hours, then perhaps log the issue as a separate one (with a question of replacing the graphing library), remove this update function, and let the tooltip for each legend option suffice for the immediate future. Particularly since the expansion of the long text was not part of this original request.

src/app/wallet/proposals/proposal-details/proposal-details.component.ts
vikas-cis-zz: yeah correct !! I missed the console :P :D

vikas-cis-zz:
@zaSmilingIdiot that's it I think what we can play with the current library attributes and if need to have different then we can log the ticket for it.

Let me know if that make sense 

Thanks

zaSmilingIdiot (approved):

PR #1477, opened 2019-03-01
Creates separation between mainnet and testnet particld binaries used

zaSmilingIdiot:
This change attempts to fix 2 things:

1. Allowing the desktop application to distinguish between mainnet and testnet particld binaries. That is, allow for different core binary versions to be used when running the application in mainnet or testnet modes.

2. The application not shutting down correctly when the daemon has not started up correctly. In cases such as when a new core version is downloading, or the daemon is not managed by the GUI, the application does not correctly shutdown and stop. Seems to be related to electron.app.quit()  never being triggered if the daemon is not running. Addition effort has been made to try and cleanly stop the download stream if a new version of core is currently running.

### DIFF
--- a/main.js
+++ b/main.js
@@ -15,7 +15,7 @@ if (process.platform === 'linux') {
 
 /* check for paths existence and create */
 [ app.getPath('userData'),
-  app.getPath('userData') + '/testnet'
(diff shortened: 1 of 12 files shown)


vikas-cis-zz:
Looks Good to me (y)

PR #1483, opened 2019-03-05
accept & terms text updated

vikas-cis-zz:
https://particl.atlassian.net/browse/PBD-71


AllienWorks:
I cleaned up some redundant tags, LGTM now!

vikas-cis-zz:
Thanks @AllienWorks

zaSmilingIdiot (changes requested):
Seems that when the modal is displayed (or very shortly after the modal is displayed) the body of text is scrolled automatically (meaning that the T&C starts +- half way down). The modal should keep the text at the top of the window until the user manually scrolls.

zaSmilingIdiot (approved):

AllienWorks:
> Seems that when the modal is displayed (or very shortly after the modal is displayed) the body of text is scrolled automatically (meaning that the T&C starts +- half way down). The modal should keep the text at the top of the window until the user manually scrolls.

For future reference, this was caused by the [modal automatically focusing first tabbable element](https://material.angular.io/components/dialog/overview#dialog-content):

> Once a dialog opens, the dialog will automatically focus the first tabbable element.
> You can control which elements are tab stops with the tabindex attribute:
> `<button mat-button tabindex="-1">Not Tabbable</button>`

zaSmilingIdiot:
Yep, thanks for pointing that out :)
I probably should have updated this PR with a reference to the one in which the correct fix was applied. So here it is :wink: 
https://github.com/particl/particl-desktop/pull/1504

PR #1485, opened 2019-03-06
Adjusts configuration for marketplace to startup correctly on mainnet

zaSmilingIdiot:
This change updates the used configuration such that the applications starts the marketplace correctly for launch on mainnet (or with any provided custom configuration).

Highlights of the change:
- Start marketplace only after the daemon has initialized;
- Merge the config options provided in particl.conf (either default or via custom datadir if a datadir cli flag is passed) and the application startup cli flags so as to correctly determine such things as the port to use. This is necessary as the particl.conf file might contain the necessary port to connect to the daemon, but the desktop application was expecting this to be passed in as a cli option. This has been updated so that the application does not need explicit flags if the particl.conf file contains the necessary settings (precedence applies, so cli flags will take precedence, etc). And also ensures that the angular app itself is not dependant on hardcoded environment values (used more as fallback default values if no other config is provided).
- Passes the config options into the angular application so that the wallet, marketplace, etc have the relevant connection details.
- Starts the marketplace with updated environment values so that it connects using the correct startup configuration.
- Fixes a test case or 2 that was preventing the tests from completing successfully.


vikas-cis-zz:
@zaSmilingIdiot I have tested on mac with different scenarios and it looks good to me 👍

PR #1501, opened 2019-03-10
update primary green in app to match new logo green

LiteBit:
updated primary green to not clash with new logo green. 

old primary green `#03e8b0`
new primary green `#45d492`

**source:** https://github.com/particl/branding/blob/master/Styleguide/particl_styleguide.pdf

_note:_ I didn't touch other gradients of the old green, only the primary green.


AllienWorks:
Added all the gajillion logos, icons and graphics based on our dev version. Should be good to go for mainnet @zaSmilingIdiot :)

zaSmilingIdiot (approved):

PR #1507, opened 2019-03-13
Fixes electron renderer restart resulting in Angular RPC calls failing.

zaSmilingIdiot:
This change ensures that the angular RpcService fetches the daemon connection configuration from the main process rather than waiting for it to simply be emitted..

This fixes such issues where, for example, development live reload or making use of the 'Close Window' window command restarts the renderer process which then waits indefinitely for RPC connection details and appears that the daemon is no longer connected/running.

### DIFF
--- a/main.js
+++ b/main.js
@@ -110,6 +110,13 @@ function initMainWindow() {
   mainWindow.setMenuBarVisibility(false);
   mainWindow.setAutoHideMenuBar(true);
 
+  // Setup configuration listener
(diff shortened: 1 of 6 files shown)


zaSmilingIdiot:
Also now includes the fix for the loading bar appearing to load infinitely on first load of the Overview page's Transaction table.

vikas-cis-zz:
https://particl.atlassian.net/browse/PBD-253 

the above changes also fixed the above task ^^

Thanks @zaSmilingIdiot

PR #1453, opened 2019-02-14
added the actual balance on overview page

vikas-cis-zz:
https://particl.atlassian.net/browse/PBD-226

PR #1516, opened 2019-03-18
GFX: Balances in sidebar

AllienWorks:
Related to #1453 and [PBD-249](https://particl.atlassian.net/browse/PBD-249), this is a proposal for always reachable overview of balances.

After much thinking and experimenting, I believe this is a way to go:

sidebar-balances

The logo section is always visible (even on smaller screens), it doesn't scroll. I decided to go only with "locked, spendable & total balances" in the end. I don't think knowing exactly how much Blind/Anon/staking balance is necessary at all times – when you're sending those, you have the balance preview on Send page anyway (in other cases you don't need them at all).

Coded only statically so far, if you agree with this, it's ready for dev implementation.

### DIFF
--- a/src/app/core-ui/main/main-view.component.html
+++ b/src/app/core-ui/main/main-view.component.html
@@ -4,6 +4,20 @@
 
     <div class="logo">
       <img src="./assets/particl-logo{{testnet ? '-testnet' : ''}}.svg" alt="Particl">
+      <table class="balances" cellspacing="0">
+        <tr>
+          <th>Spendable</th>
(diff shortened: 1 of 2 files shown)


zaSmilingIdiot:
Looks good. In my opinion the chosen values are the correct ones to display.

AllienWorks:
I think we can close this one, changes have been implemented in #1453.

zaSmilingIdiot:
As per previous comment, completed in another branch/PR

PR #1488, opened 2019-03-06
[PBD-233] MP/GUI - Multi wallet implementation (Old branch investigation)

spazzymoto:
Merged the old multi wallet branch into the current market branch.

Ensured that market only starts up and shuts down on a certain wallet.

PR #1544, opened 2019-04-01
Fixes the current wallet create/restore process

zaSmilingIdiot:
Additionally, includes select element change events that were previously broken by the dependency updates..


zaSmilingIdiot:
Closing this PR as the current workflow will include multiwallets, which has the create/restore wallet process inline rather than as modals (ie: the current create-wallet.component.ts modal is replaced with a new component in which the steps have been re-worked).

See https://github.com/particl/particl-desktop/pull/1488

PR #1565, opened 2019-04-10
Community Test Build 20190411

zaSmilingIdiot:


zaSmilingIdiot:
Leaving the comment here that this builds is meant to only include testing for the multi-wallet implementation. The application initially appears similar to the v1.x versions of the wallet. That is, the marketplace will not be visible. The market related actions are only present on a wallet named  `Market`
(note the capital letter 'M').

**IMPORTANT**: If testing this build, make certain that the `particl-market` folder does not exist on your machine. This folder needs to be removed before running this build.

Please note that the build does not include the updates to the marketplace yet, which would bring in the omp-lib integration. Hence, any known or previously reported issues in the buy flow that are meant to be resolved with the omp-lib integration will likely still be present.

zaSmilingIdiot:
Hmmm... seems to have timed out. Will start again...

zaSmilingIdiot:
Build links should no longer be available. Closing this as its essentially a snapshot of the state of the market-beta branch.

PR #1590, opened 2019-05-02
GFX: Theming tweaks

AllienWorks:
No much has changed visually, but this PR brings updates & tweaks to how our themes are handled (now properly, Material-way).

This will also help with dynamic themes in the future, which I wanted to keep away as a surprise, but you'd find out anyway, so.. :)


zaSmilingIdiot:
@AllienWorks Looks like travis ran into an issue with the tests... testing locally results in the same error.

While we can increase the amount of memory node uses to run with during the tests, which should help solving the problem, I'm not particularly keen on doing that unnecessarily. To that end, I've traced the issue to the inclusion of both light and dark themes, specifically, adding both the following imports into `_config.scss`:
```
// Import themes
@import 'theme-light';
@import 'theme-dark';
```
Notably, just including one or the other starts the tests runner at least. 
Testing this a little more specifically, removing the line:
`@include angular-material-theme($dark-theme);`
for example from either of the theme files seems to work as well (while importing both theme files in _config.scss).

Seems there is a problem with including angular-material-theme more than once? Anyways, looking at the make-up of its inclusion, the same data/inputs/construct are used in both theme files. Would it not be simpler for the sake of theme creation (and to possibly solve this issue), to take the whole of:

```
// instead of creating a theme with mat-light-theme like a sane person, we will create our own theme-creating function that lets us apply our own foreground and background palettes.
@function create-part-theme($primary, $accent, $warn) {
  @return (
    primary: $primary,
    accent: $accent,
    warn: $warn,
    is-dark: true,
    foreground: $theme-foreground,
    background: $theme-background,
  );
}

// The rest of this is like a normal theme.scss file. The only difference being that we use create-part-dark-theme instead of mat-light-theme
$theme-primary: mat-palette($part-green, 500, 100, 900);
$theme-accent: mat-palette($part-blue, 500, 100, 900);
$theme-warn: mat-palette($part-red);

$dark-theme: create-part-theme($theme-primary, $theme-accent, $theme-warn);


// Include theme styles for core and each component used in app
.part-dark-theme {
  @include angular-material-theme($dark-theme);
}
```

out of each theme file and import it once (possibly via a separate scss file??) in _config.scss after the theme files have been imported? Replacing the variables $dark-theme and $light-theme with just something like $theme. I'm not sure if its possible, so just checking...

AllienWorks:
> To that end, I've traced the issue to the inclusion of both light and dark themes, specifically, adding both the following imports into `_config.scss`

> Seems there is a problem with including angular-material-theme more than once?

This is a correct way of adding multiple themes which can be switched dynamically. Honestly I'm not quite sure what might be the problem there.

Basically you define the `primary`, `accent` and `warn` colors of your theme. You can use the Material color pallets for this or define your custom ones (what I did).

This will however only swap the main colors. As I was experimenting with dark theme, it become apparent that we'll need to do more than that, ie. redefine even the foreground and background colors for the components - that's what [`_theme-dark.scss`](https://github.com/particl/particl-desktop/pull/1590/files#diff-417ffa893988cbc49217a8bcdb3dd7d7) and [`_theme-light.scss`](https://github.com/particl/particl-desktop/pull/1590/files#diff-04f4490d7588ea7b95fc1f065477798e) do. Than you take these presets and pair them with some class for when they should be used:

```
// Include theme styles for core and each component used in app
.part-dark-theme {
  @include angular-material-theme($dark-theme);
}
```

It's way more complicated than I anticipated, as we'd need to redefine _all_ colors in _all_ components based on theme used. That would mean _a lot_ of new code for each new theme used. In my naivity, I thought one small file with defined colors for each theme would be enough, but it's not the case.

The dark theme isn't even super nice at the moment - even half-finished.. You can check it with replacing the class in `index.html` file:

```
<body class="part-light-theme">
- to -
<body class="part-dark-theme">
```

Screenshot from 2019-05-03 14-01-59

So yes, I'm having second thoughts about this. All new code will need to be styled two times (color-wise) for 2 themes. If more themes are created, it will get even worse.

At this point I'm thinking "let's stick to one default theme only". If you guys agree, I will remove the dark theme code from this PR. However this PR has even more changes which I'd like to see merged.

zaSmilingIdiot:
Fair enough... you're the expert on these things which is why I was asking. :D

I got the whole material theme setup (actually did some looking into how the themes were generated, so I felt a little more educated :b). From I could determine, most places were 
`@include angular-material-theme(...)` only once, which resulted in the previous observations when I tried removing from just one of the themes. Which led to the whole question of whether it was possible to define the themes, and then include  `angular-material-theme` once somewhere, and pass all the theme definitions to it. Rather than including it in every theme file and importing all theme files directly to it.

The way I would have seen this working is (which is subject to my own experiences so may be a little shortsighted or "not quite right", so take it for what it is): 
- the themes defined would have been pre-built (using gulp or grunt for example). This might be included as part of a pre-build or pre-compile step of the application.
- on application startup, check whether a saved theme exists. If not, use a predefined default theme file.
- using JS and referencing the index.html page, inject/replace the css src for the selected theme into the index.html page.

Granted, I haven't tried that with Material themes before, so oh well :D

Anyways...

If you're happy to keep a single theme for now then I am as well. And if there are other changes that you want included here, then perhaps removing one of themes for now is better, just to get those changes in.

AllienWorks:
OK, dark theme removed + some small tweaks added. Ready to merge @zaSmilingIdiot

PR #1519, opened 2019-03-19
Updates dependencies (develop branch)

zaSmilingIdiot:
Updates dependencies on the develop branch for the mainnet desktop wallet.

This change updates the application to make use of Angular 7.2

PR #1639, opened 2019-08-03
recovery: make transactions to stealth addresses more recoverable

kewde:
This PR will scan the blockchain for new transactions each time a stealth address is generated to make recovering the funds easier as long as there's no lookahead.


zaSmilingIdiot:
:) Cherry-picked into the 1.4.0 branch, and PR https://github.com/particl/particl-desktop/pull/1519
which still needs to be merged into develop.

zaSmilingIdiot (approved):

PR #1886, opened 2025-09-05
governance: use cryptoguard repo

nahuhh:


nahuhh:
> Merging is blocked
Commits must have verified signatures.

Well, thats too bad. GitHub wont accept my email address (which is tied to my pgp key), so i cannot make so-called "github-verified" commits, even if i gpg sign them.

